PR #1, opened 2017-04-10
Initial overview

rossberg-chromium:
This is a brain dump capturing various previous discussions and some of my notes. Lots more to be said and discussed, obviously, but we have to start somewhere.

### DIFF
--- a/proposals/gc/Overview.md
+++ b/proposals/gc/Overview.md
@@ -1,1 +1,687 @@
+# GC Extension
+
+## Introduction
+
+### Motivation
+
+* Efficient support for high-level languages
+  - faster execution
+  - smaller deliverables
+  - the vast majority of modern languages need it
+
+* Efficient interoperability with embedder heap
+  - for example, DOM objects on the web
+  - no space leaks from mapping bidirectional tables
+
+* Provide access to industrial-strength GCs
+  - at least on the web, VMs already have high performance GCs
+
+* Non-goal: seamless interoperability between multiple languages
+
+
+### Challenges
+
+* Fast but type-safe
+* Lean but sufficiently universal
+* Language-independent
+* Trade-off triangle between simplicity, expressiveness and performance
+* Interaction with threads
+
+
+### Approach
+
+* Only basic but general structure: tuples (structs) and arrays.
+* No heavyweight object model.
+* Accept minimal amount of dynamic overhead (checked casts) as price for simplicity/universality.
+* Independent from linear memory.
+* Pay as you go.
+* Avoid generics or other complex type structure if possible.
+
+
+### Requirements
+
+* Allocation of structures on the heap which are garbage collected.
+* Allocation of unstructured byte arrays which are garbage collected.
+* Handles to heap values from the embedder, garbage collected.
+Constructing closures which are  garbage collected.
+Manipulating references to these, as value types.
+Forming unions of different types, as value types.
+Defining, allocating, and indexing structures as extensions to imported types.
+Exceptions
+Direct support for strings?
+
+
+### Efficiency Considerations
+
+Managed Wasm should inherit the efficiency properties of unmanaged Wasm as much as possible, namely:
+
+* all operations are very cheap, ideally constant time,
+* structures are contiguous, dense chunks of memory,
+* accessing fields are single-indirection loads and stores,
+* allocation is fast,
+* no implicit boxing operations (i.e. no implicit allocation on the heap),
+* primitive values should not need to be boxed to be stored in managed data structures.
+
+
+## Use Cases
+
+### Tuples and Arrays
+
+* Want to represent first-class tuples/records/structs with static indexing.
+* Want to represent arrays with dynamic indexing.
+* Possibly want to create arrays with both fixed or dynamic size.
+
+Example (fictional language):
+```
+type tup = (int, int, bool)
+type vec3d = float[3]
+type buf = {pos : int, buf : char[]}
+```
+
+Needs:
+
+* user-defined structures and arrays as heap objects,
+* references to those as first-class values.
+
+
+### Objects and Method Tables
+
+* Want to represent instances as structures, whose first field is the method table.
+* Want to represent method tables themselves as structures, whose fields are function pointers.
+* Subtyping is relevant, both on instance types and method table types.
+
+Example (Java-ish):
+```
+class C {
+  int a;
+  void f(int i);
+  int g();
+}
+class D extends C {
+  double b;
+  override int g();
+  int h();
+}
+```
+
+```
+(type $f-sig (func (param (ref $C)) (param i32)))   ;; first param is `this`
+(type $g-sig (func (param (ref $C)) (result i32)))
+(type $h-sig (func (param (ref $D)) (result i32)))
+
+(type $C (struct (ref $C-vt) (mut i32))
+(type $C-vt (struct (ref $f-sig) (ref $gh-sig)))    ;; all immutable
+(type $D (struct (ref $D-vt) (mut i32) (mut f64)))  ;; subtype of $C
+(type $D-vt (struct (extend $C-vt) (ref $g-sig)))   ;; immutable, subtype of $C-vt
+```
+
+Needs:
+
+* (structural) subtyping,
+* immutable fields (for sound subtyping),
+* universal type of references,
+* down casts
+* dynamic linking might add a whole new dimension.
+
+To emulate the covariance of the `this` parameter, one down cast on `this` is needed in the compilation of each method that overrides a method from a base class.
+For example, `D.g`:
+```
+(func $D.g (param $Cthis (ref $C))
+  (local $this (ref $D))
+  (set_local $clos (call $outer (f64.const 1)))
+  (block $fail (result (ref $D))
+    (set_local $this (cast_down (ref $Cthis) (ref $D) $fail (get_local $Cthis)))
+    ...
+  )
+  (unreachable)
+)
+```
+
+
+### Closures
+
+* Want to represent closures as pairs of typed function pointers and typed environment record.
+
+Example:
+```
+function outer(x : f64) : float -> float {
+  let a = x + 1
+  function inner(y : float) {
+    return y + a + x
+  }
+  return inner
+}
+
+function caller() {
+  return outer(1)(2)
+}
+```
+
+```
+(type $func-f64-f64 (func (param $env $anyref) (param $y f64) (result f64)))
+(type $clos-f64-f64 (struct (field $code (ref $func-f64-f64)) (field $env anyref)))
+(type $inner-env (struct (field $x i32) (field $a i32)))
+
+(func $outer (param $x f64) (result (ref $clos-f64-f64))
+  (ref_func $inner)
+  (get_local $x)
+  (f64.add (get_local $x) (f64.const 1))
+  (new $inner-env)
+  (new $clos-f64-f64)
+)
+
+(func $inner (param $anyenv anyref) (param $y f64) (result f64)
+  (local $env (ref $inner-env))
+  (block $fail (result (ref anyref))
+    (set_local $env (cast_down anyref (ref $inner-env) $fail (get_local _$anyenv)))
+    (get_local $y)
+    (get_field $inner-env $a (get_local $env))
+    (f64.add)
+    (get_field $inner-env $x (get_local $env))
+    (f64.add)
+    (return)
+  )
+  (unreachable)
+)
+
+(func $caller (result f64)
+  (local $clos (ref $clos-f64-f64))
+  (set_local $clos (call $outer (f64.const 1)))
+  (call_ref
+    (get_field $clos-f64-f64 $env (get_local $clos))
+    (f64.const 2)
+    (get_field $clos-f64-f64 $code (get_local $clos))
+  )
+)
+```
+
+Needs:
+* function pointers
+* universal type of references
+* down casts
+
+The down cast for the closure environment is necessary because expressing its static type faithfully would require first-class generics and existential types.
+
+An alternative is to provide [primitive support](#closures) for closures.
+
+
+### Parametric Polymorphism
+
+TODO: via type `anyref` and `intref`
+
+
+### Type Export/Import
+
+* Want to allow type definitions to be imported from other modules.
+* As much as possible of the above constructions should be allowed with abstract types.
+* More complicated linking patterns might require user-space linking hooks.
+* Possibly: allow abstract type exports (encapsulation)?
+* Lots of tricky details here, mostly ignore for now...
+
+
+## Basic Functionality: Simple aggregates
+
+* Extend the Wasm type section with new constructors for aggregate types.
+* Extend the value types with new constructors for references and interior references.
+* Aggregate types are not value types, only references to them are.
+* References are never null; nullable reference types are separate.
+
+
+### Structs
+
+*Structure* types define aggregates with heterogeneous fields that are _statically indexed_:
+```
+(type $time (struct (field i32) (field f64)))
+(type $point (struct (field $x f64) (field $y f64) (field $z f64)))
+```
+Such types can be used by forming *reference types*, which are a new type of value type:
+```
+(local $var (ref $point))
+```
+
+Fields are *accessed* with generic load/store instructions that take a reference to a structure:
+```
+(func $f (param $p (ref $point))
+  (store_field $point $y (get_local $p)
+    (load_field $point $x (get_local $p))
+  )
+)
+```
+All accesses are type-checked at validation time.
+
+Structures are [allocated](#allocation) with `new` instructions that take initialization values for each field.
+The operator yields a reference to the respective type:
+```
+(func $g
+  (call $g (new_struct $point (i32.const 1) (i32.const 2) (i32.const 3)))
+)
+```
+Structures are garbage-collected.
+
+
+### Arrays
+
+*Array* types define aggregates with _homogeneous elements_ that are _dynamically indexed_:
+```
+(type $vector (array f64))
+(type $matrix (array (type $vector)))
+```
+Such types again can be used by forming reference types.
+For now, we assume that all array types have dynamic ([flexible](#flexible-aggregates)) size.
+
+Elements are accessed with generic load/store instructions that take a reference to an array:
+```
+(func $f (param $v (ref $vector))
+  (store_elem $vector (get_local $v) (i32.const 1)
+    (load_elem $vector (get_local $v) (i32.const 2))
+  )
+)
+```
+The element type of every access is checked at validation time.
+The index is checked at execution time.
+A trap occurs if the index is out of bounds.
+
+Arrays are [allocated](#allocation) with `new` instructions that take a size and an initialization value as operands, yielding a reference:
+```
+(func $g
+  (call $g (new_array $vector (i32.const 0) (i64.const 3.14)))
+)
+```
+Arrays are garbage-collected.
+
+The *length* of an array, i.e., the number of elements, can be inquired via the `load_length` instruction:
+```
+(load_length $vector (get_local $v))
+```
+
+
+### Packed Fields
+
+Fields and elements can have a packed *storage type* `i8` or `i16`:
+```
+(type $s (struct (field $a i8) (field $b i16)))
+(type $buf (array i8))
+```
+The order of fields is not observable, so implementations are free to optimize types by reordering fields or adding gaps for alignment.
+
+Packed fields require special load/store instructions:
+```
+(load_field_packed_s $s $a (...))
+(load_field_packed_u $s $a (...))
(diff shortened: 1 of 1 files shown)


proposals/gc/Overview.md
```diff
@@ -1 +1,696 @@
...
+* universal type of references,
+* down casts
+* dynamic linking might add a whole new dimension.
+
+To emulate the covariance of the `this` parameter, one down cast on `this` is needed in the compilation of each method that overrides a method from a base class.
+For example, `D.g`:
+```
+(func $D.g (param $Cthis (ref $C))
+  (local $this (ref $D))
+  (set_local $clos (call $outer (f64.const 1)))
```
lukewagner: Was this $clos line intended?

proposals/gc/Overview.md
```diff
@@ -1 +1,696 @@
...
+### Universal Type
+
+The type `anyref` can hold references of any reference type.
+It can be formed via [up casts](#casting),
+and the original type can be recovered via [down casts](#casting).
+
+
+### Foreign References
+
+A new built-in value type called `foreignref` represents opaque pointers to objects on the _embedder_'s heap.
```
lukewagner: On this subject, I also think we should have the ability to have a `foreignref<T>`.  This type will allow calls to imported Web API functions to also mention `foreignref<T>` in their signature and the result will be really fast, cast-free calls from wasm into Web APIs.  The "T" would just be, I think, just an arbitrary byte string and dynamic casts would be required to go between any foreign<T> to foreign<U> where T != U, and whether the cast succeeds is up to the embedder.

proposals/gc/Overview.md
```diff
@@ -1 +1,696 @@
...
+### Mutability
+
+Fields and elements can either be immutable or *mutable*:
+```
+(type $s (struct (field $a (mut i32)) (field $b i32)))
+(type $a (array (mut i32)))
+```
+Store operators are only valid when targeting a mutable field or element.
+
+Immutability is needed to enable the safe and efficient [subtyping](#subtyping), especially as needed for the [objects](#objects-and-mehtod-tables) use case.
```
bnjbvr: nit: `mehtod` => `method` in the link

proposals/gc/Overview.md
```diff
@@ -1 +1,696 @@
...
+The type `anyref` can hold references of any reference type.
+It can be formed via [up casts](#casting),
+and the original type can be recovered via [down casts](#casting).
+
+
+### Foreign References
+
+A new built-in value type called `foreignref` represents opaque pointers to objects on the _embedder_'s heap.
+
+There are no operations to manipulate foreign references, but by passing them as parameters or results of exorted Wasm functions, embedder references (such as DOM objects) can safely be stored in or round-trip through Wasm code.
```
bnjbvr: `exorted` => `exported`

proposals/gc/Overview.md
```diff
@@ -1 +1,696 @@
...
+
+Structures are [allocated](#allocation) with `new` instructions that take initialization values for each field.
+The operator yields a reference to the respective type:
+```
+(func $g
+  (call $g (new_struct $point (i32.const 1) (i32.const 2) (i32.const 3)))
+)
+```
+Structures are garbage-collected.
+
```
rossberg-chromium: Added that structures can be compared for identity.

proposals/gc/Overview.md
```diff
@@ -1 +1,696 @@
...
+Function references can be called through `call_ref` instruction:
+```
+(type $t (func (param i32))
+
+(func $f (param $x (ref $t))
+  (call_ref (i32.const 5) (get_local $x))
+)
+```
+Unlike `call_indirect`, this instruction is statically typed and does not involve any runtime check.
```
bnjbvr: Could these function references also be put in Tables (which type would be the function's type), for `call_indirect` with no runtime checks? (the current call_indirect being a regression from asm.js which only had strictly-typed tables)

proposals/gc/Overview.md
rossberg-chromium: Done.

proposals/gc/Overview.md
rossberg-chromium: Done.

proposals/gc/Overview.md
```diff
@@ -1 +1,696 @@
...
+casts the operand of type `<type1>` to type `<type2>`.
+An upcast is always safe.
+It is a validation error if the operand's type is not `<type1>`, or if `<type1>` is not a subtype of `<type2>`.
+
+Casting is also possible in the reverse direction:
+```
+(cast_down <type1> <type2> $label (...))
+```
+also casts the operand of type `<type1>` to type `<type2>`.
+It is a validation error if the operand's type is not `<type1>`, or if `<type1>` is not a subtype of `<type2>`.
```
bnjbvr: Shouldn't `type1` be a supertype of `type2`? (looks like a copy-pasto)

rossberg-chromium:
>
> Shouldn't type1 be a supertype of type2? (looks like a copy-pasto)
>
Yup, fixed.

proposals/gc/Overview.md
```diff
@@ -1 +1,716 @@
...
+```
+(func $f (param $p (ref $point))
+  (store_field $point $y (get_local $p)
+    (load_field $point $x (get_local $p))
+  )
+)
+```
+All accesses are type-checked at validation time.
+
+Structures are [allocated](#allocation) with `new` instructions that take initialization values for each field.
```
bnjbvr: (nit: `#allocation` doesn't seem to link to a section as of now)

Should we mention (somewhere, probably in the future `#allocation` paragraph) that structs/arrays might **not** get allocated/garbage-collected, if the implementation can prove it doesn't to be allocated (through escape analysis and scalar replacement)? Maybe too much of an implementation detail at the moment. The current phrasing just made me wonder if it is a strong requirement that structures **must** get allocated.

rossberg-chromium: Removed xref. Yeah, probably not worth going into details about possible optimisations at this level. As usual, implementations can do whatever is semantically equivalent. Additional language should only be necessary if we explicitly want to _forbid_ certain optimisations, e.g., like we did with TCO.

proposals/gc/Overview.md
```diff
@@ -1 +1,716 @@
...
+Fields and elements can have a packed *storage type* `i8` or `i16`:
+```
+(type $s (struct (field $a i8) (field $b i16)))
+(type $buf (array i8))
+```
+The order of fields is not observable, so implementations are free to optimize types by reordering fields or adding gaps for alignment.
+
+Packed fields require special load/store instructions:
+```
+(load_field_packed_s $s $a (...))
```
smvv: Two questions:
1. Can't the *signedness* of the field be inferred from the type descriptor?
2. Can't *packed* be inferred from the type descriptor as well?

rossberg-chromium: Wasm does not distinguish signed from unsigned integer types, so it is necessary here to specify sign extension behaviour, same as with existing instructions such as memory loads and stores. That also implies that at least the packed loads are different opcodes than non-packed loads. For symmetry and clarity it seems preferable to do the same for stores then.

smvv: Can i8 and i16 be used on non-packed structs, and if so, why is that?

Regarding signedness, let's say we export a type to the JavaScript world. If we export i8, i16, and i32 fields without any signedness information, it is impossible to determine to sign-ext or zero-ext when loading a value from that field in JS. TypedObject have u8, u16, u32 for this reason: https://github.com/nikomatsakis/typed-objects-explainer/blob/master/core.md#primitive-type-definitions How should we expose this signedness information to the JS world?

proposals/gc/Overview.md
```diff
@@ -1 +1,716 @@
...
+* Language-independent
+* Trade-off triangle between simplicity, expressiveness and performance
+* Interaction with threads
+
+
+### Approach
+
+* Only basic but general structure: tuples (structs) and arrays.
+* No heavyweight object model.
+* Accept minimal amount of dynamic overhead (checked casts) as price for simplicity/universality.
```
mtrofin: ...but that's a cost paid only if you opt-in to GC, i.e. vanilla wasm code isn't affected, correct?

proposals/gc/Overview.md
```diff
@@ -1 +1,716 @@
...
+* Independent from linear memory.
+* Pay as you go.
+* Avoid generics or other complex type structure _if possible_.
+* Extend the design iteratively, ship a minimal set of functionality fast.
+
+
+### Requirements
+
+* Allocation of structures on the heap which are garbage collected.
```
mtrofin: which heap? JS heap, or the wasm memory?

rossberg-chromium:
>
> Can i8 and i16 be used on non-packed structs, and if so, why is that?
>
There are no packed structs, only "packed" (sub-word-sized) fields. These
types can only be used for fields, they are not general value types.


> Regarding signedness, let's say we export a type to the JavaScript world.
> If we export i8, i16, and i32 fields without any signedness information, it
> is impossible to determine to sign-ext or zero-ext when loading a value
> from that field in JS. TypedObject have u8, u16, u32 for this reason:
> https://github.com/nikomatsakis/typed-objects-
> explainer/blob/master/core.md#primitive-type-definitions How should we
> expose this signedness information to the JS world?
>
You are right, that can be a hassle. However, this problem isn't new, it
already exists with plain Wasm today: whenever you import/export a function
with an integer param/result from or to JavaScript, it is the
responsibility of the JS side to interpret (and potentially, convert) their
signedness correctly. At some point we discussed adding some form of
annotation to ease that.

In general, JS interop and API is a somewhat separate problem layer that
probably needs its own subdiscussion.

proposals/gc/Overview.md
```diff
@@ -1 +1,716 @@
...
+* Direct support for strings? (separate proposal)
+
+
+### Efficiency Considerations
+
+GC support should maintain Wasm's efficiency properties as much as possible, namely:
+
+* all operations are very cheap, ideally constant time,
+* structures are contiguous, dense chunks of memory,
+* accessing fields are single-indirection loads and stores,
```
mtrofin: Won't tuple-typed fields need more than that (atomic access to avoid tearing)

proposals/gc/Overview.md
```diff
@@ -1 +1,716 @@
+# GC Extension
```
mtrofin: Is it about "GC"-ed languages or is it "JIT" rather?

proposals/gc/Overview.md
lukewagner: Perhaps we could have a new section at the end that just says "these types could be added as element types of tables, along with `get_elem`/`set_elem` ops"?

proposals/gc/Overview.md
rossberg-chromium: Not sure I understand the question. Why JIT?

proposals/gc/Overview.md
rossberg-chromium: Right. Added as comment to the pay-as-you-go bullet

proposals/gc/Overview.md
rossberg-chromium: There is no access to tuple-typed fields. With nesting, you can get an interior reference to such a field, but you can only load/store its fields individually.

But I added a bullet to requirements about interaction with threads.

proposals/gc/Overview.md
rossberg-chromium: Conceptually, the (newly introduced) Wasm heap -- which, in a JS embedding, will happen to be shared with JS.

proposals/gc/Overview.md
rossberg-chromium: Wouldn't homogeneous tables simply have a plain function type as element type?

I can see that with function references we might want to add instructions to access tables directly, but that in turn seems independent of homogeneous tables.

Trying to unravel the dependencies. It seems that homogeneous function tables are unrelated to GC types. The only feature through which tables (both homegeneous and heterogeneous) might interact with this proposal would be the instructions you mention.

proposals/gc/Overview.md
mtrofin: Not sure myself anymore either. Somehow I probably misread something, and started thinking "JIT". Sorry.

proposals/gc/Overview.md
```diff
@@ -1 +1,717 @@
...
+* no implicit boxing operations (i.e. no implicit allocation on the heap),
+* primitive values should not need to be boxed to be stored in managed data structures,
+* allows ahead-of-time compilation and code caching.
+
+
+### Evaluation
+
+Should attempt to implement 2-3 exemplary languages:
+
+* an object-oriented language (e.g., a subset of Java, with classes, inheritence, interfaces),
```
bashor: Could we help you by implementing it for Kotlin?

sjrd: I also have plans to compile Scala.js to wasm-with-GC (background info: I'm the author of Scala.js), as soon as some prototype interpreter is available. From there, it shouldn't take more than a 3-4 weeks, since we can reuse our entire pipeline except the very last piece (the so-called "emitter"), which for the compile-to-JS version is less than 4000 LoCs.

rossberg-chromium: Any implementation experiments will help evaluating the design, and any feedback from such experiments will be highly appreciated. The more the better! That said, we probably want to make sure that the set of experimental compilers includes some sufficiently "mainstream" languages of each category.

proposals/gc/Overview.md
```diff
@@ -1 +1,728 @@
...
+* a typed functional language (e.g., a subset of ML, with closures, polymorphism, variant types)
```
evancz: I'm the designer and lead developer of [Elm](http://elm-lang.org/), a simple ML-family language with all these features. Our primary compilation target is JS right now, so DOM interaction is a very practical consideration for us. Our community is very interested in WebAssembly, and I would like to help with this.

proposals/gc/Overview.md
```diff
@@ -1 +1,728 @@
...
+
+* Allocation of structures that are garbage collected.
+* Allocation of unstructured byte arrays that are garbage collected.
+* Handles to heap values from the embedder, garbage collected.
+* Manipulating references to these, as value types.
+* Forming unions of different types, as value types? (future extension?)
+* Defining, allocating, and indexing structures as extensions to imported types? (future extension)
+* Exceptions (separate proposal)
+* Direct support for strings? (separate proposal)
+* Safe interaction with threads (sharing, atomic access)
```
evancz: Does "atomic access" mean atomic *writes*? Or something more restrictive?

(For context, Elm only has immutable values, and we are quite interested in creating light-weight processes in the spirit of Erlang. Point is, it seems that cheap *reads* would be very valuable for us.)

evancz: Does "atomic access" mean atomic *writes*?

All Elm values are immutable, and we are well positioned to take advantage of threading. Point is, it seems that cheap *reads* could be quite valuable in our case.

evancz: It seems too early to pin this down, but I just want to point out that Elm only has immutable values and is well positioned to take advantage of threading quite extensively. Point is, I suspect a distinction between atomic *reads* and atomic *writes* would be valuable for us.

jfbastien: Right, it would also be nice to clarify no-tearing versus ordering guarantees.

rossberg-chromium: Agreed, but first things first. For starters, thread interaction is just a big TODO, see e.g. the section on Sharing. I anticipate more details being filled out eventually.

proposals/gc/Overview.md
rossberg-chromium: Excellent!

proposals/gc/Overview.md
jfbastien: I was replying to a comment which now seems deleted.

proposals/gc/Overview.md
```diff
@@ -1 +1,728 @@
...
+  int h();
+}
+```
+
+```
+(type $f-sig (func (param (ref $C)) (param i32)))   ;; first param is `this`
+(type $g-sig (func (param (ref $C)) (result i32)))
+(type $h-sig (func (param (ref $D)) (result i32)))
+
+(type $C (struct (ref $C-vt) (mut i32))
```
konsoletyper: In my current implementation of JVM upon linear memory I store reference to class in first 4 bytes. Besides vtable, class contains information about object's size. GC uses this information to get object's size. How GC is intended to know size of a tuple? I guess, physically every tuple will have 4 or 8 bytes header with tuple type. And *then* goes reference to vtable, which is additionally 4 or 8 bytes. Would it be useful to support special kind of first field of a tuple to implement similar behaviour?

konsoletyper: Class is not just a vtable. In my current implementation I found following fields: flags (initalized, primitive), identifier, name, class of element (for array classes), arrayof class, reference to supertype function, superclass, reference to enum fields. Do you suggest to provide this information via special function which is present in every vtable?

proposals/gc/Overview.md
```diff
@@ -1 +1,728 @@
...
+)
+```
+Structures are garbage-collected.
+
+Structures can be compared for identity:
+```
+(same (new $point ...) (new $point ...))  ;; false
+```
+TODO: Could even allow heterogeneous equality (equality between operands of different type), but that might lead to some discontinueties or even prevent some potential optimizations?
+
```
konsoletyper: What about shallow copy of a structure (which is usually implemented by memcpy)?

proposals/gc/Overview.md
```diff
@@ -1 +1,728 @@
...
+```
+Arrays are garbage-collected.
+
+The *length* of an array, i.e., the number of elements, can be inquired via the `load_length` instruction:
+```
+(load_length $vector (get_local $v))
+```
+
+Like structures, arrays can be compared for identity.
+
```
konsoletyper: In JVM, arrays are first-class objects that extend `java.lang.Object`, support `getClass()`, `hashCode`, etc. In my implementations arrays are actually objects that have corresponding header, which points to class with vtable. How I can implement this with Wasm arrays? The only approach I can see is to generate additional wrapper class (two structures, one for data and one for vtable).

proposals/gc/Overview.md
```diff
@@ -1 +1,728 @@
...
+Note: Like [type equivalence](#type-equivalence), subtyping is *structural*.
+The above is the standard (co-inductive) definition, which is the most general definition that is sound.
+Checking it is computationally equivalent to checking whether one FSA recognises a sublanguage of another FSA, i.e., it is a non-trivial algorithm (even though most practical cases will be trivial).
+Like with type equivalence, this may be a problem, in which case a more restrictive definition might be needed.
+
+Subtyping could be relaxed such that mutable fields/elements could be subtypes of immutable ones.
+That would simplify creation of immutable objects, by first creating them as mutable, initialize them, and then cast away their constness.
+On the other hand, it means that immutable fields can still change, preventing various access optimizations.
+(Another alternative would be a three-state mutability algebra.)
+
```
konsoletyper: So, types are organized in a tree, which is insufficient for Java, where types form DAG. How are interfaces supposed to be implemented?

proposals/gc/Overview.md
konsoletyper: Ah, I just realized that this "vtable" structure is not necessarily a structure of function pointers. It can contain all these fields as-is.

rossberg-chromium:
>
> In my current implementation of JVM upon linear memory I store reference
> to class in first 4 bytes. Besides vtable, class contains information about
> object's size. GC uses this information to get object's size. How GC is
> intended to know size of a tuple? I guess, physically every tuple will have
> 4 or 8 bytes header with tuple type. And *then* goes reference to vtable,
> which is additionally 4 or 8 bytes. Would it be useful to support special
> kind of first field of a tuple to implement similar behaviour?
>
@konsoletyper, yes, this will most likely result in taking up two words for
each Java object. Coming up with a feature for Wasm that would make this
more space efficient while still being portable across all engines and
hardware platforms is a serious challenge. If anybody knows a satisfactory
solution to that problem then I'd be very interested in seeing it.

proposals/gc/Overview.md
rossberg-chromium: Right. Not currently included, but may be a worthwhile extension if there turn out to be sufficiently many use cases of large structures that need to be copied by value.

proposals/gc/Overview.md
rossberg-chromium: With the base proposal you indeed need to allocate two values, the "object" and the array backing store. With the extension for nesting and "flexible types" discussed below you could embed the latter into the former.

proposals/gc/Overview.md
rossberg-chromium: No, structural subtyping can form arbitrary DAGs. Also, note that these are low-level representation types. They don't necessarily bear a direct relation to the source language type system.

rossberg-chromium:
Is anybody opposed to landing this initial PR?

PR #3, opened 2017-04-25
Add interpreter support for GC objects

smvv:
This PR implements the NEW_OBJECT, STORE_FIELD and LOAD_FIELD opcode for object support.

```
$ cat ~/work/tests/wasm-object-support/load_field.wast
;; vim: ft=lisp ts=2
(module
  (type $s
    (struct (field f64) (field f64))
    )
  (memory 1)
  (func
    (param f64 f64)
    (result (ref $s))
    (local (ref $s))
    new_object (ref $s)

    tee_local 2
    get_local 0
    store_field $s 0

    get_local 2
    get_local 1
    store_field $s 1

    get_local 2
    return
    )
  (func
    (export "main")
    (param f64 f64)
    (result f64)
    (local (ref $s))
    get_local 0
    get_local 1
    call 0

    tee_local 2
    load_field $s 0
    get_local 2
    load_field $s 1
    f64.add
    )
  )

(assert_return (invoke "main" (f64.const 1337) (f64.const 42)) (f64.const 1379))
$ ./wasm -t ~/work/tests/wasm-object-support/load_field.wast
-- Running ("(input \"/home/smvv/work/tests/wasm-object-support/load_field.wast\")")...
-- Parsing...
-- Running...
-- Loading (/home/smvv/work/tests/wasm-object-support/load_field.wast)...
-- Parsing...
-- Running...
-- Checking...
-- Initializing...
-- Asserting return...
-- Invoking function "main"...
GetLocal x=0 v=[]
GetLocal x=1 v=1337.
NewObject x=0 v=[]
obj={type_index:0,fields:[0.,0.]}
TeeLocal x=2 v={type_index:0,fields:[0.,0.]}
GetLocal x=0 v={type_index:0,fields:[0.,0.]}
StoreField vs=[{type_index:0,fields:[0.,0.]} 1337.] struct=0 field=0
obj={type_index:0,fields:[1337.,0.]}
GetLocal x=2 v=[]
GetLocal x=1 v={type_index:0,fields:[1337.,0.]}
StoreField vs=[{type_index:0,fields:[1337.,0.]} 42.] struct=0 field=1
obj={type_index:0,fields:[1337.,42.]}
GetLocal x=2 v=[]
TeeLocal x=2 v={type_index:0,fields:[1337.,42.]}
LoadField vs={type_index:0,fields:[1337.,42.]} struct=0 field=0
obj={type_index:0,fields:[1337.,42.]}
GetLocal x=2 v=1337.
LoadField vs=[1337. {type_index:0,fields:[1337.,42.]}] struct=0 field=1
obj={type_index:0,fields:[1337.,42.]}
```

There are some TODO items left though:

- [ ] Fix grammar issue with `value_type_list` and `block` and `if`. The optional value type list that is returned from a block/if starts with an opening parenthesis, and an instruction following the optional value type list can also start with an opening parenthesis. This results in shift/reduce conflicts.
  **Question:** Is adding the `LPAR TYPE value_type_list RPAR` instead of `value_type_list`
  a solution to this problem? See the changes to `IF` and `block:` in parser.mly.

- [ ] Structs declared after a function cannot be referenced in that function in the WAST syntax. Struct declarations should only happen in the first section ("types") to avoid looking up a type index of a struct that is not yet declared. In the binary format, it is not possible to have this issue since all types are stored in the section `types` which is the first section.
  **Question:** Should this limitation be part of the Wast grammar? Currently it is possible to have types after functions causing runtime errors.

- [ ] Rename TypeDescrType to StructType, and rename TypeDescrElemType to StructElemType. Fix the 'TypeDescrElemType TypeDescrType' match statement in `spec/valid.ml` as well. This is easy to solve but I've postponed it since other problems where more important to solve first.

- [ ] Remove all printf calls (currently used for debugging purposes).

- [ ] Implement DOWN_CAST and UP_CAST opcodes.


interpreter/spec/types.ml
```diff
@@ -1,9 +1,15 @@
 (* Types *)
 
 type value_type = I32Type | I64Type | F32Type | F64Type
+    | ObjType of int32
 type elem_type = AnyFuncType
 type stack_type = value_type list
 type func_type = FuncType of stack_type * stack_type
+type type_descr_type = TypeDescrType of stack_type
```
rossberg-chromium: Call this `struct_type`?

interpreter/spec/eval.ml
```diff
@@ -398,8 +453,11 @@ let check_limits actual expected at =
 let add_import (ext : extern) (im : import) (inst : instance) : instance =
   let {ikind; _} = im.it in
   match ext, ikind.it with
-  | ExternalFunc clos, FuncImport x when func_type_of clos = type_ inst x ->
-    {inst with funcs = clos :: inst.funcs}
+  | ExternalFunc clos, FuncImport x
+    when func_type_of clos = match type_ inst x with
```
rossberg-chromium: This occurs repeatedly. How about a helper `func_type : inst -> var -> func_type`?

interpreter/spec/eval_numeric.ml
```diff
@@ -181,17 +181,36 @@ struct
     | DemoteF64 -> raise (TypeError (1, v, F64Type))
 end
 
+module ObjOp =
```
rossberg-chromium: This shouldn't be needed if you refine the type algebra.

interpreter/spec/types.ml
```diff
@@ -1,9 +1,15 @@
...
+
+type func_or_type_descr_type =
```
rossberg-chromium: `def_type`?

interpreter/spec/valid.ml
```diff
@@ -231,6 +239,18 @@ let rec check_instr (c : context) (e : instr) (s : infer_stack_type) : op_type =
     require (mut = Mutable) x.at "global is immutable";
     [t] --> []
 
+  | LoadField ({struct_ = _; field = _} as fo) ->
```
rossberg-chromium: You don't need to match the record if you're not binding any variable inside.

interpreter/spec/values.ml
```diff
@@ -3,10 +3,48 @@ open Types
 
 (* Values and operators *)
 
-type ('i32, 'i64, 'f32, 'f64) op =
-  I32 of 'i32 | I64 of 'i64 | F32 of 'f32 | F64 of 'f64
+type ('i32, 'i64, 'f32, 'f64, 'obj) op =
```
rossberg-chromium: This shouldn't include reference types.

interpreter/spec/values.ml
```diff
@@ -3,10 +3,48 @@ open Types
...
+  I32 of 'i32 | I64 of 'i64 | F32 of 'f32 | F64 of 'f64 | Obj of obj_type
+and value = (I32.t, I64.t, F32.t, F64.t, obj_type) op
+and obj_type = {
+    type_index : int32;
+    fields : value array option;
```
rossberg-chromium: There shouldn't be any uninitialized fields in the semantics, so you should be able to get by without an option here.

interpreter/spec/ast.ml
```diff
@@ -66,6 +75,9 @@ type storeop = Memory.mem_size memop
 type var = int32 Source.phrase
 type literal = Values.value Source.phrase
 
+type loadfieldop = {struct_ : var; field : int32}
```
rossberg-chromium: No particular reason to separate load and store version of these. Also, `field` should probably be a `var`, too.

interpreter/text/parser.mly
```diff
@@ -198,8 +206,14 @@ text_list :
...
-  | /* empty */ { [] }
-  | VALUE_TYPE value_type_list { $1 :: $2 }
+  | /* empty */ { fun c -> [] }
+  | VALUE_TYPE value_type_list { fun c -> $1 :: ($2 c) }
+/* TODO:
+  | LPAR REF NAT RPAR
+    { fun c -> ObjType ($3 c type_).it }
+*/
+  | LPAR REF VAR RPAR value_type_list
+    { fun c -> ObjType (type_ c ($3 @@ at ())) :: ($5 c) }
```
rossberg-chromium: Hm, the AST should reflect the input format, so in the syntax, a reference type should just contain a `var`. That would probably also solve your forward reference problem and avoid all the extra thunkification you introduced. And it might be necessary to support recursive types anyway. However, it might also mean that you need to pass in the type context in a couple of places.

interpreter/text/parser.mly
```diff
@@ -579,11 +609,23 @@ inline_export :
 
 /* Modules */
 
+struct_fields :
+  | /* empty */ { fun c -> [] }
+  | LPAR FIELD VALUE_TYPE RPAR struct_fields
```
rossberg-chromium: Are you planning to add symbolic names for field indices?

interpreter/text/parser.mly
```diff
@@ -347,9 +367,12 @@ expr1 :  /* Sugar */
     { fun c -> let c' = $2 c [] in let ts, es = $3 c' in [], block ts es }
   | LOOP labeling_opt block
     { fun c -> let c' = $2 c [] in let ts, es = $3 c' in [], loop ts es }
-  | IF labeling_opt value_type_list if_
+  | IF labeling_opt if_
     { fun c -> let c' = $2 c [] in
-      let es, es1, es2 = $4 c c' in es, if_ $3 es1 es2 }
+      let es, es1, es2 = $3 c c' in es, if_ [] es1 es2 }
+  | IF labeling_opt LPAR TYPE value_type_list RPAR if_
```
rossberg-chromium: `type` isn't the right keyword here, but I wanted to propose changing the syntax of block types anyway, in order to be future compatible with generalised block types. With that, result types need to be enclosed in `(result ...)`, analogous to functions. That might solve your problem as well.

Issue #10, opened 2017-07-13
Eliminate additional allocation in closure example


wingo:
Hi!  Really great proposal.  I had a look at it from the point of view of whether it's suitable as a target for a Scheme compiler, and generally, really good stuff.  I had a few thoughts and will file specific issues.

This issue relates to the closure example.  I think you probably agree that we should aim to have examples that are minimal, illustrative, and optimal.  That way they can be easy to understand while also illustrating that the proposal is capable of translating optimal examples.

In that regard I think we could revise the closure example to avoid the double allocation, basically putting the function pointer in with the data.  Keeping the same example:

```
function outer(x : f64) : float -> float {
  let a = x + 1
  function inner(y : float) {
    return y + a + x
  }
  return inner
}

function caller() {
  return outer(1)(2)
}
```

```
(type $clos-func-f64-f64 (func (param $env $clos-f64-f64) (param $y f64) (result f64)))
(type $clos-f64-f64 (struct (field $code (ref $clos-func-f64-f64)))
(type $inner-clos (struct (extend $clos-f46-f64) (field $x i32) (field $a i32))

(func $outer (param $x f64) (result (ref $clos-f64-f64))
  (ref_func $inner)
  (get_local $x)
  (f64.add (get_local $x) (f64.const 1))
  (new $clos-f64-f64)
)

(func $inner (param $clos (ref $clos-f64-f64)) (param $y f64) (result f64)
  (local $inner-clos (ref $inner-clos))
  (block $fail (result (ref $clos-f64-f64))
    (set_local $inner-clos (cast_down (ref $clos-f64-f64) (ref $inner-clos) $fail (get_local $clos)))
    (get_local $y)
    (get_field $inner-clos $a (get_local $inner-clos))
    (f64.add)
    (get_field $inner-clos $x (get_local $inner-clos))
    (f64.add)
    (return)
  )
  (unreachable)
)

(func $caller (result f64)
  (local $clos (ref $clos-f64-f64))
  (set_local $clos (call $outer (f64.const 1)))
  (call_ref
    (get-local $clos)
    (f64.const 2)
    (get_field $clos-f64-f64 $code (get_local $clos))
  )
)
```

[...]

Issue #7, opened 2017-06-29
Please provide some way to create encapsulated closures


erights:
It says that a closure is a pair of code and state. Is this pair separable? For code given just the pair, can it access the state other than by invoking that closure's code on that state?

If the answer is yes, then could we have encapsulating closures, i.e., invocable inseparable pairs of code and state?

rossberg-chromium:
The bare encoding given in the overview could be encapsulated if we introduced abstract types. If made abstract, the type of each closure environment would not be known to other modules, and would thus be entirely opaque. Without that, the encoding would not be encapsulated.

FWIW, the same applies to the usual representation of objects.

For closures, another option would be to introduce closures as a primitive notion, as briefly hinted at in the overview.

erights:
Thanks. I now understand the answer to my question, and so have renamed this issue into a feature request.

tlively:
I believe this is the goal of the `func.bind` instruction mentioned in the [typed function references](https://github.com/WebAssembly/function-references) proposal. I'll close this old issue in favor of continued discussion on that repo.

rossberg:
As a side note: I think we more or less converged on dropping func.bind from the function references proposal. I plan to remove it there and move it to the Post-MVP doc here.

PR #18, opened 2017-07-13
Eliminate second allocation in closure example

wingo:
Also, remove the future idea of primitive closure support.  Choosing
the representation of a closure is part of the domain of the compiler
that targets WebAssembly; WebAssembly can just provide the primitives to
support the choices that a compiler would like to make.

### DIFF
--- a/proposals/gc/Overview.md
+++ b/proposals/gc/Overview.md
@@ -151,7 +151,8 @@ For example, `D.g`:
 
 ### Closures
 
-* Want to represent closures as pairs of typed function pointers and typed environment record.
+* Want to associate a code pointer and its "environment" in a GC-managed object
+* Want to be able to allow compiler of source language to choose appropriate environment representation
 
 Example:
 ```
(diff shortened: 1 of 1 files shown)


wingo:
Fixes #10.

proposals/gc/Overview.md
```diff
@@ -169,26 +170,25 @@ function caller() {
...
-  (new $inner-env)
   (new $clos-f64-f64)
 )
 
-(func $inner (param $anyenv anyref) (param $y f64) (result f64)
-  (local $env (ref $inner-env))
-  (block $fail (result (ref anyref))
-    (set_local $env (cast_down anyref (ref $inner-env) $fail (get_local _$anyenv)))
+(func $inner (param $clos (ref $clos-f64-f64)) (param $y f64) (result f64)
+  (local $inner-clos (ref $inner-clos))
```
rossberg-chromium: Nit: `$clos` would seem good enough, since there is only one closure (and avoids confusion with the type name).

proposals/gc/Overview.md
```diff
@@ -579,11 +580,6 @@ TODO: The ability to import types makes the type and import sections interdepend
 TODO
 
 
-## Possible Extension: Closures
```
rossberg-chromium: Please don't remove prematurely. There are a number reasons why we still might want closures as a primitive mechanism, see e.g. #7, or language interop, or more efficient support for currying, etc.

proposals/gc/Overview.md
```diff
@@ -208,12 +208,13 @@ function caller() {
 
 Needs:
 * function pointers
-* universal type of references
+* recursive types
```
rossberg-chromium: Nit: more specifically, (mutually) recursive function types.

wingo:
Thanks for the review.  Fixed nits, I think; PTAL :)

rossberg-chromium:
Thanks!

proposals/gc/Overview.md
wingo: `$clos` is also the formal parameter name so it's unavailable.  Still, I changed the local name to `$env`; at least it's shorter and doesn't collide with the type name.  Happy to change to something else if you think of a better name.

PR #20, opened 2017-07-17
Minor fixes and rewordings

titzer:
Made a pass through the overview section, removing the use of "tuple" and using "length" for arrays instead of "size", as well as small fixes elsewhere.

### DIFF
--- a/proposals/gc/Overview.md
+++ b/proposals/gc/Overview.md
@@ -6,10 +6,10 @@
 
 * Efficient support for high-level languages
   - faster execution
-  - smaller deliverables
+  - smaller modules
   - the vast majority of modern languages need it
 
-* Efficient interoperability with embedder heap
+* Efficient interoperability with embedder
   - for example, DOM objects on the web
   - no space leaks from mapping bidirectional tables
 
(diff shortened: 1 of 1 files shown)


proposals/gc/Overview.md
```diff
@@ -297,20 +300,19 @@ Elements are accessed with generic load/store instructions that take a reference
...
   (call $g (new_array $vector (i32.const 0) (i64.const 3.14)))
 )
 ```
-Arrays are garbage-collected.
 
-The *length* of an array, i.e., the number of elements, can be inquired via the `load_length` instruction:
+The *length* of an array, i.e., the number of elements, can be inquired via the `load_array_length` instruction:
 ```
-(load_length $vector (get_local $v))
+(load_array_length $vector (get_local $v))
```
rossberg-chromium: Nit: inserting "array" here is inconsistent with the naming used for the other instructions, which don't have that either.

proposals/gc/Overview.md
```diff
@@ -323,7 +325,7 @@ Fields and elements can have a packed *storage type* `i8` or `i16`:
 (type $s (struct (field $a i8) (field $b i16)))
 (type $buf (array i8))
 ```
-The order of fields is not observable, so implementations are free to optimize types by reordering fields or adding gaps for alignment.
+The order of fields is not observable, so implementations are free to reorganize the underlying memory for their storage or add alignment.
```
rossberg-chromium: Nit: "add gaps for alignment"? Because alignment itself isn't "added".

proposals/gc/Overview.md
```diff
@@ -574,7 +579,7 @@ Imported types are essentially parameters to the module.
 As such, they are entirely abstract, as far as compile-time validation is concerned.
 The only operations possible with them are those that do not require knowledge of their actual definition or size: primarily, passing and storing references to such types.
 
-TODO: The ability to import types makes the type and import sections interdependent.
+TODO: The ability to import types makes the type and import sections interdependent. (we will probably need to add placeholder entries in the types section for imported types, and possibly may need to add assumptions about their representation, even to compile code that uses them as locals, unless we want to restrict imported types to be only reference types).
```
rossberg-chromium: The assumptions could in fact be expressed as ordinary supertype bounds.

Not sure I follow the placeholder idea.

proposals/gc/Overview.md
```diff
@@ -30,58 +30,61 @@
...
-* No heavyweight object model.
-* Accept minimal amount of dynamic overhead (checked casts) as price for simplicity/universality.
-* Independent from linear memory.
-* Pay as you go; in particular, no effect on code not using GC.
-* Don't introduce dependencies on GC for other features (e.g., using resources through tables).
-* Avoid generics or other complex type structure _if possible_.
-* Extend the design iteratively, ship a minimal set of functionality fast.
+* Only basic but general structure: structs and arrays
+* No heavyweight object model
+* Accept minimal amount of dynamic overead (checked casts) as price for simplicity/universality
```
titzer: Must have been fat-fingered. Fixed

proposals/gc/Overview.md
```diff
@@ -30,58 +30,61 @@
...
 * a typed functional language (e.g., a subset of ML, with closures, polymorphism, variant types)
 * an untyped language (e.g., a subset of Scheme or Python or something else)
 
 
 ## Use Cases
 
-### Tuples and Arrays
+### Structs and Arrays
 
-* Want to represent first-class tuples/records/structs with static indexing.
```
titzer: Done.

proposals/gc/Overview.md
```diff
@@ -211,11 +214,11 @@ Needs:
 * (mutually) recursive function types
 * down casts
 
-The down cast for the closure environment is necessary because expressing its static type faithfully would require first-class generics and existential types.
+The down cast for the closure environment is necessary because expressing a static type system would require first-class generics and existential types.
```
titzer: Well, let's not get into the semantics of the word "express", but I've reworded a bit.

proposals/gc/Overview.md
```diff
@@ -225,19 +228,19 @@ TODO: via type `anyref` and `intref`
...
 * Lots of tricky details here, mostly ignore for now...
 
 
 ## Basic Functionality: Simple Aggregates
 
-* Extend the Wasm type section with new constructors for aggregate types.
-* Extend the value types with new constructors for references and interior references.
-* Aggregate types are not value types, only references to them are.
-* References are never null; nullable reference types are separate.
+* Extend the Wasm type section with new operators to allocate aggregate types
```
titzer: Ah yes, peephole was a bit too small during this pass.

proposals/gc/Overview.md
```diff
@@ -275,7 +278,7 @@ Structures can be compared for identity:
 ```
 (same (new $point ...) (new $point ...))  ;; false
 ```
-TODO: Could even allow heterogeneous equality (equality between operands of different type), but that might lead to some discontinueties or even prevent some potential optimizations?
+TODO: Could even allow equality comparisons between operands of different type, but that might lead to some discontinuities or even prevent some potential optimizations?
```
titzer: Argh, this is a TODO so we shouldn't bikeshed it to death, but in this case I found that the term to add more cognitive overhead than it saved, since it immediately had to be summarize with a parenthetical anyway.

proposals/gc/Overview.md
titzer: This is to make it explicit, since other structures might have lengths in the future. "array_length" is a thing in the "verb_thing" sense, not a "verb_type_thing" sense.

proposals/gc/Overview.md
titzer: Reworded.

proposals/gc/Overview.md
titzer: The placeholder idea was to reserve indices in the type space for imported types, but I guess we could do the same thing as function types and prepend the imported types to the type index space. I suppose the constraints could be expressed in the import.

Note that supertype bounds aren't really enough; one could view imported types as a poor man's type parameters, and generating code that manipulates imported types (even just for calling conventions) would require knowledge of the representation (thus the comment about assuming imports can only be reference types).

rossberg-chromium: Wouldn't an upper bound like `anyref` express that knowledge? Wasm types mostly are representation types, after all.

proposals/gc/Overview.md
rossberg-chromium: Well, it's concise and convenient to use in discussions, so introducing it should help.

rossberg:
Sorry, I somehow dropped the ball on this entirely.

PR #24, opened 2017-10-12
Fix duplicated 'add's in README.md

honggyukim:
Hi, I just found a mistake on README.md while just passing by so fixed it.

Thanks,
Honggyu


rossberg-chromium:
Thanks!

PR #33, opened 2018-03-23
Correct a few typos

bnjbvr:
Also removes the Closures TODO section, since there's another one above in the
document.


bnjbvr:
@rossberg PTAL.

PR #38, opened 2018-10-07
Add explicit RTT support for casts

rossberg:
This came out of the discussion on PR #34.

Also, add more motivating explanation in overview.

### DIFF
--- a/proposals/gc/MVP.md
+++ b/proposals/gc/MVP.md
@@ -27,6 +27,10 @@ Based on [reference types proposal](https://github.com/WebAssembly/reference-typ
 * `i31ref` is a new reference type
   - `reftype ::= ... | i31ref`
 
+* `rtt <reftype>` is a new reference type that is a runtime representation of type `<reftype>`
+  - `reftype ::= ... | rtt <reftype>`
+  - `rtt t ok` iff `t ok`
+
 
 #### Type Definitions
 
@@ -53,8 +57,11 @@ Based on [reference types proposal](https://github.com/WebAssembly/reference-typ
 #### Imports
 
 * `type <typetype>` is an import description with an upper bound
-  - `importdesc ::= ... | type <reftype>`
+  - `importdesc ::= ... | type <typetype>`
   - Note: `type` may get additional parameters in the future
+
+* `typetype` describes the type of a type import, and is either an upper bound or a type equivalence
+  - `typetype ::= sub <reftype> | eq <reftype>`
 
 * Type imports have indices prepended to the type index space, similar to other imports.
   - Note: due to bounds, type imports can be mutually recursive with other type imports as well as regular type definitions. Hence they have to be validated together with the type section.
@@ -75,7 +82,7 @@ In addition to the rules for basic reference types:
 
 * `eqref` is a subtype of `anyref`
   - `eqref <: anyref`
-  - Note: `i31ref` and `anyfunc` are *not* a subtypes of `eqref`, i.e., those types do not expose reference equality
+  - Note: `i31ref` and `funcref` are *not* a subtypes of `eqref`, i.e., those types do not expose reference equality
 
 * `nullref` is a subtype of `eqref`
   - `nullref <: eqref`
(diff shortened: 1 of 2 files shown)


proposals/gc/Overview.md
```diff
@@ -231,7 +242,76 @@ An alternative is to provide [primitive support](#closures) for closures, e.g. a
...
+
+function g(p1 : (Bool, Bool), p2 : (C, C), pick : <A> ((A, A)) -> A) : C {
+  if (pick<Bool>(p1))
+    return pick<C>(p2);
+  else
+    return new C;
+}
+```
+
+Here, `make_pair` as well as `fst` and `snd` need to be able to operate on any type of pair. Furthermore, `fst` and `snd` cannot simply be type-specialised at compile time, because that would be insufficient to compile `g`, which takes a polymorphic function as an argument and instantiates it with multiple different types. Such *first-class* polymorphism is not expressible with compile time techniques such as C++ templates, but common-place in many languages (including OO ones like Java or C#, where it can be emulated via generic methods). Untyped languages like JavaScript or Scheme trivially allow such programs as well.
```
aardappel: I think this could be explained more precisely. In particular, there will be programming languages for which the type checker is able to determine the return type of the use of `pick` in `g`, and thus can compile it to safe code that does not need a `cast_down` (e.g. when translated directly to LLVM, C or x86). So the need for a RTT / `cast_down` is not just about first class polymorphism, but it is about a mismatch between the language and what wasm types can express (since we apparently don't want to have the concept of statically typed generics at the wasm level). This a tradeoff we're choosing, i.e. we could instead decide to make wasm types a whole lot more complicated in exchange for needing less or no `cast_down` ops or need for a RTT field in all "objects".

This is similar to what Java does, which I am not a fan of, though I understand why dealing with generics at the wasm level may not be impractical. On a related note, letting inefficiencies like this creep into the format we're encouraging engines to do their own type analysis to omit `cast_down` and RTT fields where possible (it wouldn't surprise me if Hotspot does this already).

rossberg: @aardappel, I agree that this is a choice we make. I added a couple more sentences mentioning the possible future addition of type parameters to Wasm to avoid this overhead.

To be clear, the problem isn't the ability to deduce the return type -- AFAICT, all languages are able to do that (that's the point of generics after all). The problem is that the compilation of `fst` and `snd` itself must not depend on that type because you generally cannot tell, at compile time, the set of all types they will be instantiated with (static analysis can do that in many cases but not all). So compile-time type specialisation does not generally work with first-class polymorphism. Unless you are willing to implement runtime compilation and type specialisation (like C# / .NET) you have to rely on a type-agnostic compilation scheme, and short of having type parameters in Wasm, going to the top type and casting back is the only alternative available.

Added a version of this as explanation.

aardappel: Ah, good point about `fst` and `snd`, thanks. Good to hear you think type parameters could still be added in the future.

proposals/gc/MVP.md
```diff
@@ -122,6 +129,10 @@ In addition to the rules for basic reference types:
   - `var <valtype> <: var <valtype>`
   - Note: mutable fields are *not* subtypes of immutable ones, so `const` really means constant, not read-only
 
+* `rtt t` is a subtype of `anyref`
+  - `rtt t <: anyref`
+  - Note: `rtt t1` is *not* a subtype of `rtt t2`, even if `t1` is a subtype of `t2`; such subtyping would be unsound, since RTTs are used in both co- and contravariant roles
```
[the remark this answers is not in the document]
rossberg: Done.

proposals/gc/Overview.md
```diff
@@ -550,29 +638,51 @@ On the other hand, it means that immutable fields can still change, preventing v
...
+
+In order to cast down the type of a struct or array, the aggregate itself must also be equipped with a suitable RTT. Attaching runtime type information to aggregates happens at allocation time but is optional. If no RTT is attached then their runtime type is treated as if it was `anyref` and a down cast to a more specific type will fail. Such aggregates can prevent client code from rediscovering their real type, enforcing a form of parametricity. They can also be optimised more aggressively (e.g., via flattening optimisations), since the VM knows that any possible additional fields forgotten via subtyping can never be rediscovered.
 
-Casting is also possible in the reverse direction:
+A runtime type is an expression of type `rtt <type>`, which is another form of opaque reference type. It denotes the static type `<type>`.
+
+Runtime type checks verify the subtype relation between runtime types.
+In order to make these checks cheap, runtime subtyping follows a *nominal* semantics. To that and, every RTT value not only represents a given type, it also records a subtype relation to another (runtime) type (possibly `anyref`) defined when constructing the RTT value:
```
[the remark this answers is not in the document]
rossberg: Fixed.

PR #40, opened 2018-10-15
Additional conversion instructions

rossberg:
This adds:

* the `ref.test` instruction that tests a reference's RTT without actually casting;
* the `optref.as_nonnull` and `br_on_null` instructions that null-check and convert an optref to a ref. This is no longer subsumed by casts, because they cannot be used if the object has no RTT attached;
* the `rtt.i31ref` instruction, which was missing before;
* more compact shorthands `ref.is_func`, `ref.as_func`, `ref.is_i31`, `ref.as_i31`.

I'm somewhat undecided on the shorthands, but wanted to throw them out there for discussion.

### DIFF
--- a/proposals/gc/MVP.md
+++ b/proposals/gc/MVP.md
@@ -147,8 +147,8 @@ In addition to the rules for basic reference types:
 
 ### Values
 
-* Each reference value has an associated runtime type:
-  - For structures or arrays, it is the RTT value given upon creation, or `anyref` if none.
+* Each reference value has an associated *runtime type*, which is a runtime description of its type:
+  - For structures or arrays, it is determined by an [RTT](#runtime-types) value provided upon creation, or `anyref` if none.
   - For `i31ref` references it is `i31ref`.
   - For `null` it is `nullref`.
 
@@ -301,7 +301,7 @@ TODO: Is 31 bit value range the right choice?
 * `rtt.new <reftype> <reftype>` returns the RTT of the specified type as a subtype of a given RTT operand
   - `rtt.new t t' : [(rtt t')] -> [(rtt t)]`
     - iff `t <: t'`
-  - multiple invocations of this instruction yield fresh RTTs
+  - multiple invocations of this instruction with the same operand yield the same RTTs
 
 * All RTT instructions are considered *constant expressions*.
 
--- a/proposals/gc/MVP.md
+++ b/proposals/gc/MVP.md
@@ -31,6 +31,9 @@ Based on [reference types proposal](https://github.com/WebAssembly/reference-typ
   - `reftype ::= ... | rtt <reftype>`
   - `rtt t ok` iff `t ok`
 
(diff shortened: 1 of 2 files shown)


proposals/gc/MVP.md
```diff
@@ -266,10 +308,15 @@ TODO: Is 31 bit value range the right choice?
 
 #### Casts
 
-* `cast <reftype> <reftype>` casts a reference value down to a type given by a RTT representation
-  - `cast t t' : [t (rtt t')] -> [t']`
+* `ref.test <reftype> <reftype>` tests whether a reference value is of a type given by a RTT representation
```
lukewagner: So here we have all 3 of: test-producing-i32, trapping-cast, branching-cast, but in the other cases (i32ref, optref, funcref) we only have different subsets of these.  Is that intentional or can we have the same 3 operations for all these basic reference types?

proposals/gc/MVP.md
```diff
@@ -266,10 +308,15 @@ TODO: Is 31 bit value range the right choice?
```
lukewagner: Pre-existing, but: the statement "All RTT instructions are considered constant expressions" seems to be in conflict with `rtt.new`'s "multiple invocations of this instruction yield fresh RTTs".  Is that a typo or am I misinterpreting "constant expressions"?

proposals/gc/MVP.md
```diff
@@ -136,15 +136,23 @@ In addition to the rules for basic reference types:
...
 
 * Locals must have a type that is defaultable.
 
 * Table definitions with non-zero minimum size must have an element type that is defaultable. (Imports are not affected.)
 
 
+### Values
+
+* Each reference value has an associated runtime type:
+  - For structures or arrays, it is the RTT value given upon creation, or `anyref` if none.
```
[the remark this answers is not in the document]
rossberg: Tried to clarify.

proposals/gc/MVP.md
rossberg: For optref you actually have all three as well (ref.is_null preexists) -- optref is a bit special, since you want to be able to do these ops even when the type does not have RTT, so the RTT-based ops cannot be used.

For the other two, test and cast are merely shorthands for the generic ones that seem like they would come in handy. We could remove them. Inversely, we could also add a branching version for each, but that would be a much more heavyweight addition, so perhaps not justified as a mere shorthand.

But I don't have strong opinions either way.

proposals/gc/MVP.md
rossberg: As per the discussion at the Lyon meeting, I resolved this to make rtt.new non-generative for now.

proposals/gc/MVP.md
lukewagner: Not really, and it seems like we could switch later; it just seems preferable to have regularity between all the conceptual downcasts.

rossberg: Makes sense. I changed the text to list them as possibilities.

PR #45, opened 2018-11-06
Describe JS API for GC

lukewagner:
This PR moves the stub "JS API" section out of MVP.md into a new MVP-JS.md that defines the JS API in terms of the [JS Typed Objects proposal](https://github.com/tschneidereit/proposal-typed-objects/blob/master/explainer.md).

### DIFF
--- a/proposals/gc/MVP-JS.md
+++ b/proposals/gc/MVP-JS.md
@@ -1,0 +1,532 @@
+# GC v1 JS API
+
+This document describes the proposed changes to the
+[WebAssembly JS API spec](http://webassembly.github.io/spec/js-api/) that are
+associated with the [changes to the WebAssembly core spec](MVP.md).
+
+
+## Goals
+
+With the [Reference Types proposal](https://github.com/WebAssembly/reference-types/blob/master/proposals/reference-types/Overview.md),
+any JS value can be converted to an `anyref` value and back, preserving the
+exact JS value. The only way to create an `anyref` value in wasm is `ref.null`
+which can be trivially converted to a JS `null` ([or `undefined`](https://github.com/WebAssembly/reference-types/issues/9))
+value at the JS/wasm boundary.
+
+With the [GC proposal](Overview.md), which extends the Reference Types proposal,
+the WebAssembly JS API must now specify:
+* How do struct/array references created by [`struct.new`](https://github.com/WebAssembly/gc/blob/master/proposals/gc/MVP.md#structures)
+  and [`array.new`](https://github.com/WebAssembly/gc/blob/master/proposals/gc/MVP.md#arrays)
+  get [converted to JS values](https://webassembly.github.io/spec/js-api/index.html#tojsvalue)?
+* Which JS values can be [converted into WebAssembly values](https://webassembly.github.io/spec/js-api/index.html#towebassemblyvalue)
+  of struct/array reference type?
+
+Since normal JavaScript objects have a dynamic shape, making them incompatible
+with the wasm GC proposal's structs and arrays, it may be tempting to simply
+treat WebAssembly GC things as separate from JavaScript GC things and, e.g.,
+only allow opaque references to go back and forth (i.e., `anyref` in both
+directions).
+
+Instead, we propose a design that improves the performance of JavaScript both in
+isolation and when interoperating with WebAssembly. Specifically, this proposal
+defines WebAssembly structs/arrays to simply **be** certain special JavaScript
+objects in specification *and* presumed implementation. This design both avoids
+costly wrapper-based JS/WebAssembly interoperability schemes and also gives
+JavaScript a new kind of high-performance object that should be independently
+valuable for optimizing pure JavaScript applications.
+
+The challenge of course is to do this in a way that doesn't inject JS
+peculiarities into the core design of GC in WebAssembly or impact the efficiency
+of WebAssembly GC (even in JavaScript Host Embeddings).
+
+
+## Walkthrough
+
+The JavaScript spec allows [Exotic Objects](https://tc39.github.io/ecma262/#sec-exotic-object)
+to diverge from Ordinary Object behavior by overriding
+[spec-internal methods and slots](https://tc39.github.io/ecma262/#sec-built-in-exotic-object-internal-methods-and-slots).
+An example of this is Typed Arrays, which override various
+[internal property-access methods](https://tc39.github.io/ecma262/#sec-integer-indexed-exotic-objects)
+to turn normal JS property accesses into pure byte-array operations.
+This presently allows JavaScript and WebAssembly to both have high-performance
+access to WebAssembly's [linear memory](https://webassembly.github.io/spec/core/intro/overview.html#memory), without injecting JS
+peculiarities into the design of linear memory. We propose to do likewise for
+WebAssembly structs and arrays.
+
+To do this, we propose to revive, refurbish and repurpose the
+[old Typed Objects proposal](http://smallcultfollowing.com/babysteps/pubs/2014.04.01-TypedObjects.pdf).
+The [new Typed Objects proposal](https://github.com/tschneidereit/proposal-typed-objects/blob/master/explainer.md)
+is designed with WebAssembly in mind and is motivated as follows:
+
+
+### Value Type objects
+
+Corresponding to the set of [field types](https://github.com/WebAssembly/gc/blob/master/proposals/gc/MVP.md#type-definitions)
+in the WebAssembly GC proposal, the JS Typed Objects proposal includes an
+open-ended set of [Value Type objects](https://github.com/tschneidereit/proposal-typed-objects/blob/master/explainer.md#value-types)
+which contain `[[Read]]` and `[[Write]]` [internal methods](https://tc39.github.io/ecma262/#sec-object-internal-methods-and-internal-slots)
+that convert to and from arbitrary JavaScript values and the pure typed values
+that WebAssembly is defined to interact with.
+
+The JS Typed Object proposal includes a set of [common Value Type objects](https://github.com/tschneidereit/proposal-typed-objects/blob/master/explainer.md#value-types).
+This general set is extended with additional Value Type objects in the
+`WebAssembly` namespace [described below](#other-webassembly-value-types), so
+that all WebAssembly field types have a corresponding Value Type object. The 
+generally-applicable Value Type objects defined by the JS Typed Object proposal
+are broken into two categories:
+
+* *Primitive* Value Types, exposed as singleton objects: `uint8`,
+  `uint16`, `uint32`, `uint64`, `int8`, `int16`, `int32`, `int64`, `float32`,
+  `float64`, `any`, `string`.
+
+* *Compound* Value Types, which are produced via the `ref` property
+   on another kind of object...
+
+
+### Type Definition objects
+
+Corresponding to the [Type Definitions](https://github.com/WebAssembly/gc/blob/master/proposals/gc/MVP.md#type-definitions)
+in the WebAssembly GC proposal, the JS Typed Objects proposal defines constructors
+for [Type Definition objects](https://github.com/tschneidereit/proposal-typed-objects/blob/master/explainer.md#type-definitions).
+Type Definition objects are constructed by the [`StructType`](https://github.com/tschneidereit/proposal-typed-objects/blob/master/explainer.md#struct-type-definitions)
+and [`ArrayType`](TODO) constructors which are passed Value Type objects for
+each field/element.
+
+Type Definition objects serve three purposes:
+
+* When a type definition is [exported](https://github.com/WebAssembly/gc/blob/master/proposals/gc/MVP.md#exports)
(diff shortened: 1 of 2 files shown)


proposals/gc/MVP-JS.md
```diff
@@ -0,0 +1,532 @@
...
+
+
+### Typed objects
+
+Corresponding to the [structs](https://github.com/WebAssembly/gc/blob/master/proposals/gc/MVP.md#structures)
+and [arrays](https://github.com/WebAssembly/gc/blob/master/proposals/gc/MVP.md#arrays)
+of the WebAssembly GC proposal, the JS Typed Object proposal defines a new kind of Exotic Object
+called a [typed object](https://github.com/tschneidereit/proposal-typed-objects/blob/master/explainer.md#instantiation).
+Typed objects can be both created from JS by calling a Type Definition object
+as a constructor and created wasm from wasm by executing [`struct.new`](https://github.com/WebAssembly/gc/blob/master/proposals/gc/MVP.md#structures)
```
chicoxyzzy: "created wasm from wasm" should be "created from wasm"?

lukewagner: Thanks! Fixed.

proposals/gc/MVP-JS.md
```diff
@@ -0,0 +1,532 @@
...
+exact Type Definition object and use that type import in all relevant
+function signatures:
+
+```wat
+;; example4.wat --> example4.wasm
+(module
+    (type $Point (import "" "Point") (eq (struct (field $x i32) (field $y i32))))
+    (type $Rect (struct (field (ref $Point)) (field (ref $Point))))
+    (func (export "goWild") (param (ref $Rect))
+        (struct.set $R $x1
```
rossberg: Do you mean $Rect and $x here?

proposals/gc/MVP-JS.md
```diff
@@ -0,0 +1,532 @@
...
+WebAssembly.instantiateStreaming(fetch('example4.wasm'), {'':{Point}})
+.then(({instance}) => {
+    var rect = new Rect(new Point(1,2), new Point(3,4));
+    instance.exports.goWild(rect);  // succeeds
+    assert(rect.x1 instanceof Point);
+    assert(rect.x1.x === 10);
+    assert(rect.x1.y === 20);
+
+    const {Point2:Point, Rect2:Rect} = makeTypes();
+    instance.exports.goWild(new Rect2(new Point2(1,2), new Point2(3,4)));  // throws
```
rossberg: You might want to give a bit more explanation where exactly this would throw -- notably it's the call attempt itself that throws (in the ToWebAssemblyValue coercion), not the struct.set, if I understand correctly.

proposals/gc/MVP-JS.md
```diff
@@ -0,0 +1,532 @@
...
+this instantiation would succeed:
+
+```js
+const Point = new StructType([{type: int32}, {type: int32}]);
+const Rect = new StructType([{type: Point.ref}, {type: Point.ref}]);
+
+WebAssembly.instantiateStreaming(fetch('example5.wasm'), {'':{Point, Rect}});
+```
+
+while this instantiation would fail because of the `Point1`/`Point2` mismatch in the
```
rossberg: One interesting question here is how these additional link-time type checks are integrated into the semantics. Currently, all link-time checking is defined by the core spec. IIUC, this semantics would require the JS API to define an additional layer of checking that is a refinement of that. Is that correct?

proposals/gc/MVP-JS.md
```diff
@@ -0,0 +1,532 @@
...
+In addition to `WebAssembly.ref(T)`, these additional Value Type object singletons
+and constructor functions would be added to the `WebAssembly` namespace to
+reflect all the other field types in the GC proposal:
+* `WebAssembly.funcref` : constructor function reflecting [`anyfunc`/`funcref`](https://github.com/WebAssembly/reference-types/blob/master/proposals/reference-types/Overview.md#language-extensions).
+* `WebAssembly.eqref` : singleton reflecting [`eqref`](https://github.com/WebAssembly/gc/blob/master/proposals/gc/MVP.md#value-types)
+  whose `[[Write]]` accepts only objects, symbols and `null`.
+* `WebAssembly.optref(T)` : constructor function reflecting [`optref <typeidx>`](https://github.com/WebAssembly/gc/blob/master/proposals/gc/MVP.md#value-types),
+  for a given type definition `T`.
+* `WebAssembly.i31ref` : singleton reflecting [`i31ref`](https://github.com/WebAssembly/gc/blob/master/proposals/gc/MVP.md#value-types)
+  whose `[[Write]]` only accepts integral `Number` and `BigInt` values in the range [-2<sup>31</sup>, 2<sup>31</sup>-1].
```
rossberg: I believe negative values could only go down to -2^30.

proposals/gc/MVP-JS.md
lukewagner: You're right I did mean $Rect instead of $R, and but I did mean $x1, I just forgot to add this field name to $Rect's first field :)

proposals/gc/MVP-JS.md
lukewagner: Correct and good point, will do.

proposals/gc/MVP-JS.md
lukewagner: Correct; the JS API would have to "traverse" the types of signatures and make sure the type imports' identities all line up.  I'll add a note.

proposals/gc/MVP-JS.md
lukewagner: D'oh!

lukewagner:
Anyone else want to comment before merging?  (Say, @titzer or @ajklein?)  Of course the entire design here is still tentative (like MVP.md) and I added a note to the top saying as much, so we can revisit these choices after merging as we get some more experience using/implementing.

proposals/gc/MVP-JS.md
```diff
@@ -0,0 +1,537 @@
...
+The JS Typed Object proposal includes a set of [common Value Type objects](https://github.com/tschneidereit/proposal-typed-objects/blob/master/explainer.md#value-types).
+This general set is extended with additional Value Type objects in the
+`WebAssembly` namespace [described below](#other-webassembly-value-types), so
+that all WebAssembly field types have a corresponding Value Type object. The 
+generally-applicable Value Type objects defined by the JS Typed Object proposal
+are broken into two categories:
+
+* *Primitive* Value Types, exposed as singleton objects: `uint8`,
+  `uint16`, `uint32`, `uint64`, `int8`, `int16`, `int32`, `int64`, `float32`,
+  `float64`, `any`, `string`.
```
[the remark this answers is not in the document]
chicoxyzzy: There is indeed `string` Value Type in [Typed Objects proposal](https://github.com/tschneidereit/proposal-typed-objects/blob/master/explainer.md#value-types) and there is also `object` Value Type which is missing from Value Types list here. Both `string` and `object` are confusing.

titzer:
I've read most of this PR but I need a day or two to chew on the part about nominal types and importing.

proposals/gc/MVP-JS.md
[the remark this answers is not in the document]
lukewagner: The idea is that these Value Type objects are defined as part of the general "Typed Objects" JS feature and so the meaning of these value types is entirely JS-specific.  To access objects with these field types, wasm would have to import these types.  Static wasm type checking + dynamic coercions at the wasm/JS boundary would ensure that `string` values were always, in fact, JS strings, and `object` values were always, in fact, JS objects (i.e., `typeof(v) === "object" && v !== null`).  At instantiation time, this could eliminate what would otherwise be the runtime checks required when passing `anyref` to methods taking strings/objects.

Does that make sense?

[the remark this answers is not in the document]
chicoxyzzy: It does.

`object` value is still missing from list of Value Types here though.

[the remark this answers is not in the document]
lukewagner: Oops, that was an unintentional omission.  Fixed and thanks!

proposals/gc/MVP-JS.md
```diff
@@ -0,0 +1,538 @@
...
+
+If the JS `ref` Value Types were unified with the WebAssembly
+[`ref` type constructor](https://github.com/WebAssembly/gc/blob/master/proposals/gc/MVP.md#value-types),
+the answer would be "no": `wasmInstance` could assign any typed object with two
+`i32` fields, including those with a `null` `__proto__` and thus no `x` and `y`
+property names. This would be surprising from a JS perspective and would also
+give up future WebAssembly Host Binding optimizations that are mentioned
+at the end of this section.
+
+To avoid problem, as a starting point, the following WebAssembly module:
```
titzer: s/avoid/avoid the/

proposals/gc/MVP-JS.md
```diff
@@ -0,0 +1,538 @@
...
+Thus, the expected nominal invariants of JS can be preserved by a combination of
+instantiation-time and run-time checks.
+
+But what about the other direction: what if WebAssembly exports a struct
+type definition that uses a `ref` field type? To reflect WebAssembly structural
+subtyping, we need new Value Type objects. Since the set of Value Type objects
+is open-ended, the WebAssembly JS API adds a new Value Type constructor
+function: `WebAssembly.ref(T)`. Using this, JS can create a suitable Type
+Definition object whose instances can be passed to the above `example3.wasm`
+because its fields are structurally typed:
```
titzer: I don't quite understand this. Structural subtyping is a property of types, not fields.
Is it that case that `WebAssembly.ref(T)` constructs a JS Value Type that, when used as a field type, allows any structurally-compatible JS Value Type (and also WebAssembly type)? Afaict there can only be one such type (the principal type), so I assume that WebAssembly.ref(T) can actually return the same underlying value type for many such invocations (including WebAssembly.ref(F) that happens to be compatible?

lukewagner: Great questions:
* You're right that structural subtyping is a property of types, not fields, and while `WebAssembly.ref(T)` is only demonstrated to be used for field types, Value Type objects can also be used (as described in the later section "Other WebAssembly uses of Value Type objects") globals, tables and (as mentioned in the TODO section) typed function params/results.
* You're right that `WebAssembly.ref(T)` is not generative and will return equivalent underlying value types for equivalent arguments.  This equivalence is demonstrated in the example where `makeTypes()` uses `WebAssembly.ref(Point)`, which is ultimately called twice, producing equivalent value types.

proposals/gc/MVP-JS.md
lukewagner: Thanks, fixed.

proposals/gc/MVP-JS.md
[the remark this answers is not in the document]
lukewagner: Resolving, but feel free to reopen if this wasn't answered.

proposals/gc/MVP-JS.md
```diff
@@ -0,0 +1,538 @@
...
+the expectation is that practical uses will generate all these Type Definition
+objects as part of the usual JS boilerplate glue code and end users would simply
+interact with objects that were either tuple-y or had property names as expected
+from the source language. (For example, a WebAssembly-friendly [Flow](https://flow.org)
+variant might use structural typing for fields with [Tuple Type](https://flow.org/en/docs/types/tuples/)
+and nominal typing for fields with [Class Type](https://flow.org/en/docs/types/classes/).)
+
+In addition to the dynamic checks necessary when JS passes a value to
+WebAssembly (either as an argument to a wasm exported function or a return value
+from a wasm imported function), the instantiation algorithm must ensure that
```
lars-t-hansen: Or through a global variable or via a slot in a table with a nontrivial ref type?

lukewagner: Good point; resolving by adding "etc" linked down to the last section which specifically calls out these two cases.

proposals/gc/MVP-JS.md
```diff
@@ -0,0 +1,538 @@
...
+```js
+function makeTypes() {
+    const Point = new StructType([{name: "x", type: int32}, {name: "y", type: int32}]);
+    const Rect = new StructType([{name: "p1", type: Point.ref}, {name: "p2", type: Point.ref}]);
+    return {Point, Rect};
+}
+
+const {Point, Rect} = makeTypes();
+
+WebAssembly.instantiateStreaming(fetch('example4.wasm'), {'':{Point}})
```
jacobp100: I could be wrong here, but is this a typo?

```suggestion
WebAssembly.instantiateStreaming(fetch('example4.wasm'), {'':{Point, Rect}})
```

lukewagner: It's been a bit since I wrote this, so I could be wrong too, but I think it's correct as written: `example4.wat` only has 1 import, `Point`; the `$Rect` type is an internal definition, not an import.

titzer:
Ok, I've looked through this again and refreshed myself on it. I think it is sufficiently baked that we should land and iterate on it, particularly around the nominal/structural subtyping issues. I don't see any reason to hold it up for that.

So, lgtm, and sorry for the long delay on this.

lukewagner:
Thanks for the looking through this.  Agreed we'll want to continue iterating on the structural/nominal design.

PR #47, opened 2018-12-03
Distinguish typeuse from reftype

rossberg:
Instead of using `<reftype>` as immediates, which was a bit of a categroy error, introduce `typeuse`, which denotes either a user-defined type (via a type index) or a predefined type (via direct opcode).


rossberg:
Not sure if "beautiful" is the word invented for this... :)

Issue #85, opened 2020-03-26
exnrefs are nominal


RossTate:
`exnref`s are garbage-collected references that you cast to specific cases, so why aren't they using the same structural design as all other garbage-collected references? They're just structs with a tag field (with a hidden trace field?), so it seems inconsistent to design them differently than all the other structs with a tag field out there.

rossberg:
The type `exnref` is one built-in type. Exceptions define _values_ in this type, so don't relate to structural vs nominal typing.

It is true that exception tags are generative, however, as in any standard semantics, AFAICT. That's mainly for disambiguation, not encapsulation: unlike regular values, raised exceptions can bypass module/abstraction boundaries implicitly, and mix with others in the same type, so that you cannot generally tell what code they originate from when caught. It is essential for correctness that you can tell apart your exceptions from somebody else's exceptions.

RossTate:
So what you're saying is that, in any place where `anyref` is actually (dynamically) representing references from many languages, then (like with `exnref`) it is essential for correctness that you can tell your values apart from somebody else's references (in order to avoid unintended coincidences). Thus in such uses of `anyref` we should expect all references to be generatively tagged values of the form `mytag myref`. For the sake of discussion, let's call references following this convention `mixedref` (without actually making that a new static type).

We've also said that any coordinated interlanguage communication should be expected to be done using Interface Types. So with those two expectations, is there any situation where `anyref` is currently expected to be used that wouldn't be better served by `mixedref` or `myref` (where `myref` is whatever type is guaranteed by a module's own `mytag`)?

rossberg:
> So what you're saying is that, in any place where anyref is actually (dynamically) representing references from many languages, then (like with exnref) it is essential for correctness that you can tell your values apart from somebody else's references (in order to avoid unintended coincidences).

That's explicitly not what I am saying. I am saying that this applies to the exception type specifically, because throwing shortcuts explicit data flow through abstraction boundaries. In all regular cases, programs have a choice to not mix values in a single type, but with exceptions they haven't, because all exceptions share the same data flow path.

Moreover, a local exception may propagate through code that does not know about it, e.g., when library A invokes a higher-order function from library B, and passes one of its own function that throws an exception local to A. Interface types cannot help with such tunnelling.

[...]

PR #86, opened 2020-03-28
Incorporate generative RTT

RossTate:
Thanks to a number of discussions, I now have enough understanding of the expected conventions for this proposal to suggest a modest revision that should address a number of open questions.

First, the proposal already has in place an expectation to use run-time type information, RTT(I)s. One thing that was unclear is what precisely the semantics of RTTIs should be. For example, if $t1$ is a subtype of $t2, should an `anyref` with the RTT `(rtt.get $t1)` be castable to a `$t2` via `(rtt.get $t2)`? According to the current semantics, the answer is no&mdash;the cast should only succeed if the two types are exactly the same. This is good for performance, but it did bring into question how modules should share values via `anyref` since it means they have to agree exactly on what run-time types to use.

The expectation now, though, is that coordinated "share-nothing" module communication will be done via Interface Types. Furthermore, in #85, we agreed that for uncoordinated/heterogeneous communication (say via `anyref`), it is important for correctness that modules be able to distinguish their values (via generative tags, like with exceptions) from those of other values (even if they happen to have the same structure).

Furthermore, in WebAssembly/proposal-type-imports#7, we came to the conclusion that, in order to prevent client modules from inspecting the internals of abstract exported types, we need to be able to guarantee that exported values can only be cast to a target (structural) type if the module also exports its own (generative) `rtt` for the target type.

Altogether, this indicates that `rtt`s should be generative. On top of that, I was able to go through the overview and easily switch all the examples to generative RTTs, so the adjustment seems to be compatible with the known usage patterns. The adjustment required little change to the instructions, with the most significant being that `rtt.sub` semantics are now to always generate a new `rtt` rather than to always reproduce the same `rtt` with the same inputs.

My hope is that, by pinning answers to some of the major open questions, this gets us all on the same page as to what this proposal entails and how it should be expected to be used and implemented.

### DIFF
--- a/proposals/gc/MVP.md
+++ b/proposals/gc/MVP.md
@@ -148,9 +148,9 @@ In addition to the rules for [basic](https://github.com/WebAssembly/reference-ty
 
 * There is a runtime subtyping hierarchy on RTTs; creating an RTT requires providing a *parent type* in the form of an existing RTT; the RTT for `anyref` is the root of this hierarchy.
 
-* An RTT t1 is a *sub-RTT* of another RTT t2 iff either of the following holds:
-  - t1 and t2 represent the same static type, or
-  - t1 has a parent that is a sub-RTT of t2.
+* An RTT rtt1 is a *sub-RTT* of another RTT rtt2 iff either of the following holds:
+  - rtt1 and rtt2 are the same value, or
+  - rtt1 has a parent that is a sub-RTT of rtt2.
 
 * Validation requires that each parent type is a representative of a static supertype of its child; runtime subtyping hence is a sub-relation of static subtyping (a graph with fewer nodes and edges).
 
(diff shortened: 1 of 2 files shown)


tlively:
Can you briefly explain what "generative" means in this context and what alternatives it differentiates from?

proposals/gc/MVP.md
```diff
@@ -285,16 +277,20 @@ Perhaps also the following short-hands:
...
+  - each invocation of this instruction yields a distinct RTT
+  - equivalent to `(rtt.sub t (rtt.none))`
   - this is a *constant instruction*
-  - equivalent to `(rtt.sub t (rtt.get anyref))`, except when `t` itself is `anyref`
 
 * `rtt.sub <typeuse>` returns the RTT of the specified type as a sub-RTT of a given parent RTT operand
   - `rtt.sub t : [(rtt t')] -> [(rtt t)]`
     - iff `t <: t'`
-  - multiple invocations of this instruction with the same operand yield the same observable RTTs
+  - each invocation of this instruction yields a distinct RTT
```
RossTate: @tlively: Here is the most important change (along with the matching change in `rtt.get` to `rtt.new`). An "applicative" system returns the same value when given the same inputs (previous semantics), whereas a "generative" system returns a new value distinct from all previous values (new semantics).

proposals/gc/MVP.md
```diff
@@ -285,16 +277,20 @@ Perhaps also the following short-hands:
...
   - this is a *constant instruction*
```
RossTate: Also, this assertion was already present but wasn't true. For an applicative semantics, one must canonicalize the structure of the type (which is not constant time), hash it (again not constant time), and then use a host-global concurrent hash map to get the canonical `rtt` for the type (requiring another non-constant-time equality comparison). When the type is known it compile time, which is *currently* the case for this proposal, then all of that can be done at compile time and the instruction replaced with the precomputed `rtt`, which is where the assertion that this instruction is constant time comes from. However, the proposal also recommends adding parametric polymorphism later on, which will introduce type parameters that the type in `rtt.get` can refer to, in which case all of this will have to be done at run time and the instruction will no longer be constant time. With the generative semantics, this is constant time even in the presence of type variables.

rossberg:
The current docs do not incorporate generative types yet, neither statically nor dynamically. If we add nominal types then of course we'd also want to add generative RTTs. However, that does not eliminate the need for non-generative RTTs, for the same reason as with static types.

So such a change should _add_ `rtt.new`, not replace `rtt.get`. FWIW, having both was in my slides on an alternative design with mere tags (where they were called `tag.new` and `tag.canon`) that I didn't get to show at the meeting (but you saw the slides, I think).

proposals/gc/MVP.md
rossberg: "Constant instruction" is [spec terminology](https://webassembly.github.io/spec/core/valid/instructions.html#constant-expressions) that has nothing to do with constant time.

RossTate: Ah, thanks for the correction and the link!

RossTate:
> However, that does not eliminate the need for non-generative RTTs, for the same reason as with static types.

Could you give some examples? In your talk at the in-person meeting, you advocating for removing anything from the MVP that could reasonably be expressed without it. I removed `rtt.get` because every example could reasonably be expressed without it. (The same was not true for `rtt.new` due to the rationale above, particularly WebAssembly/proposal-type-imports#7.)

tlively:
I'll close this PR since we won't be including RTTs of any kind in the MVP.

PR #150, opened 2020-10-09
Simplify RTTs and decompose casts

### LINKED ISSUE #140: Will the restriction about global initializers be lifted?
Hello,
currently, global initializers cannot refer to locally defined globals. However, the first code block [here](https://github.com/WebAssembly/gc/blob/master/proposals/gc/MVP.md#runtime-types) implies that this restriction will be lifted or relaxed. Is it safe to assume that global initializers will be able to refer to already defined local globals?

### LINKED ISSUE #141: Should there be a constant i31ref initializer expression?
Following up on observations in https://github.com/WebAssembly/gc/issues/124#issuecomment-698283270, I am wondering whether it'd make sense to introduce a constant initializer expression for `i31ref` globals, because currently there is no trivial way to declare an `i31ref` typed global, requiring workarounds like

```wat
(module
 (import "env" "trivial_i31ref" (global $trivial_i31ref i31ref))
 (global $global_i31ref (mut i31ref) (global.get $trivial_i31ref))
)
```

Suitable solutions from simplest to most general seem to be (1) the introduction of `i31.zero`

```wat
(module
 (global $global_i31ref (mut i31ref) (i31.zero))
)
```

or (2) a variant of `i31.new` taking an immediate instead of an expression

```wat
(module
 (global $global_i31ref (mut i31ref) (i31.new_with_value value=0))
)
```

or, radically, (3) allowing global initializers to be defaultable and making `i31ref` default to contain value `0`

```wat
(module
 (global $global_i31ref (mut i31ref) (default))
)
```

The latter is more complicated but leads to potential code size savings, e.g. in the best case

```wat
(module
 (global $global_f64 (mut f64) (f64.const 0) ;; 9 bytes initializer
 (global $global_f64 (mut f64) (default) ;; 1 byte initializer
)
```

further leading to (4) allowing to omit global initializers entirely, like on locals

```wat
 (global $global_f64 (mut f64)) ;; 0 byte initializer
```

Given how fundamental the problem seems, I might as well be missing something?

rossberg:
Following a recent discussion with @jakobkummerow and @tebbi, this PR suggests a simplification to the semantics of RTTs and casts: instead of representing arbitrary heap types, RTTs only denote *concrete* types, making casts simpler and more uniform.

Previously, an RTT could represent an arbitrary heap type, including abstract types like `anyref` or `funcref`. Such RTTs could also be used as targets of a cast. Consequently, casts were a complex operation that could involve multiple steps, the number and details of which differ by source/target type. For example, to cast from `anyref` to `(ref $t)`, where `$t` is a struct, the operational semantics of `ref.cast` would need to

(1) check that the reference isn’t null or a scalar;
(2) check that it points to a struct;
(3) do the actual check on the RTTs involved.

In contrast, a cast from, say, `anyref` to abstract `funcref` involves only the first two of these steps. A cast from `funcref` to `(ref $t)` OTOH only the latter. Consequently, the operational semantics (and costs_ of a cast were dependent on the types involved, and a consumer would need to distinguish various different combinations of cases.

This PR decomposes the complex cast operation into simpler, lower-level primitives, and allows moving the case distinction to the producer, as is adequate for an “assembly” language.

Concretely, with this change,

* an RTT always denotes a *concrete* type, i.e., it’s form is simplified from `(rtt <heaptype>)` to `(rtt <typeidx>)`;

* accordingly, casts always target concrete types; moreover, they can only be applied to reference types that are statically known to carry an RTT (structs/arrays and funcs);

* instructions like `ref.is_func/as_func` become independent operations, not just shorthands for casts;

* along these lines, a new abstract type `dataref` is introduced as the common supertype of compound data (structs & arrays), which carry an actual RTT and can be the source of casts, along with respective instructions `ref.is_data/as_data`.

The net effect is that RTT-based casts are simplified to performing step (3) above, while (1/2) are performed by separate “classification” instructions that distinguish abstract heap types.

The change also clarifies possible questions about the `externref` type, which now is naturally excluded from being the target of a cast.

Other changes:

* The depth count <n> on RTTs only counts _concrete_ supertypes now, as the abstract ones are no longer relevant for casts; this also makes it agnostic to possible future refinements in the abstract type hierarchy.

* Furthermore, the depth count becomes optional, allowing to abstract over the depth of an inheritance chain where desired. When present, the static depth allows for slightly more efficient casts, otherwise it has to be loaded dynamically. (It remains to be seen how much of a performance benefit a static depth actually is. We could drop it entirely if it doesn’t turn out to matter.)

* In order to avoid double testing in cases where classifications are expected to be fallible, introduce branching versions of `ref.is/as` instructions.

* Relax the use of `global.get` in constant expressions (fixes #140).

* Define `i31.new` to be a constant expression  (fixes #141).

* Add some missing binary encodings.

### DIFF
--- a/proposals/gc/MVP.md
+++ b/proposals/gc/MVP.md
@@ -445,7 +445,7 @@ The opcode for heap types is encoded as an `s33`.
 | -0x11  | `extern`        |            | from funcref proposal |
 | -0x12  | `any`           |            | |
 | -0x13  | `eq`            |            | |
-| -0x17  | `i31`           |            | |
+| -0x16  | `i31`           |            | |
 
 #### Defined Types
 
--- a/proposals/gc/MVP.md
+++ b/proposals/gc/MVP.md
@@ -145,7 +145,7 @@ Its interpretation is defined by the host environment.
 It may contain additional host-defined types that are neither of the above three leaf type categories.
 It may also overlap with some or all of these categories, as would be observable by applying a classification instruction like `ref.is_func` to a value of type `externref`.
 The possible outcomes of such an operation hence depend on the host environment.
-(For example, in a JavaScript embedding, `externref` could be inhabited by all JS values -- which is a natural choice, because JavaScript is untyped; but some of its values are JS-side representations of Wasm values per the JS API, and those can also be observed as `data` or `func` references. Another possible interpretation could be that `data` is disjoint from `extern`, which would be determined by the coercions allowed by the JS API at the JS/Wasm boudary. While such an interpretation is probably not attractive for JavaScript, it would be natural in other embeddings such as the C/C++ API, where different references are represented with different host types.)
+(For example, in a JavaScript embedding, `externref` could be inhabited by all JS values -- which is a natural choice, because JavaScript is untyped; but some of its values are JS-side representations of Wasm values per the JS API, and those can also be observed as `data` or `func` references. Another possible interpretation could be that `data` is disjoint from `extern`, which would be determined by the coercions allowed by the JS API at the JS/Wasm boundary. While such an interpretation is probably not attractive for JavaScript, it would be natural in other embeddings such as the C/C++ API, where different references are represented with different host types.)
 
 Note: In the future, this hierarchy could be refined to distinguish compound data types that are not subtypes of `eq`.
 
@@ -196,8 +196,8 @@ Note: In the future, this hierarchy could be refined to distinguish compound dat
 * At the same time, runtime subtyping forms a linear hierarchy such that the relation can be checked efficiently using standard implementation techniques (the runtime subtype hierarchy is a tree-shaped graph).
 
 Note: RTT values correspond to type descriptors or "shape" objects as they exist in various engines. Moreover, runtime casts along the hierachy encoded in these values can be implemented in an engine efficiently by using well-known techniques such as including a vector of its (direct and indirect) super-RTTs in each RTT value (with itself as the last entry). The value `<n>` then denotes the length of this vector. A subtype check between two RTT values can be implemented as follows using such a representation. Assume RTT value v1 has static type `(rtt n1? $t1)` and v2 has type `(rtt n2? $t2)`. To check whether v1 denotes a sub-RTT of v2, first verify that `n1 >= n2` -- if both `n1` and `n2` are known statically, this can be performed at compile time; if either is not statically known, it has to be read from the respective RTT value dynamically, and `n1 >= n2` becomes a dynamic check. Then compare v2 to the n2-th entry in v1's supertype vector. If they are equal, v1 is a sub-RTT.
-In the case of actual casts, the static type of RTT v1 (taken from the  to cast) is not known at compile time, so `n1` is dynamic as well.
-(Note that `$t2` and `$t2` are not relevant for the dynamic semantics, but merely for validation.)
+In the case of actual casts, the static type of RTT v1 (obtained from the value to cast) is not known at compile time, so `n1` is dynamic as well.
+(Note that `$t1` and `$t2` are not relevant for the dynamic semantics, but merely for validation.)
 
 Example: Consider three types and corresponding RTTs:
 ```
@@ -350,9 +350,9 @@ Tentatively, support a type of guaranteed unboxed scalars.
   - traps if reference is not an integer
   - equivalent to `(block $l (param anyref) (result i31ref) (br_on_i31 $l) (unreachable))`
 
-Note: There are no instructions to check for `externref`, since that can consist of a diverse set of different object representations that would be costly to check for exhaustively.
+Note: The [reference types](https://github.com/WebAssembly/reference-types) and [typed function references](https://github.com/WebAssembly/function-references)already introduce similar `ref.is_null` and `br_on_null` instructions.
 
-TODO: Should we add `br_on_null` for completeness?
+Note: There are no instructions to check for `externref`, since that can consist of a diverse set of different object representations that would be costly to check for exhaustively.
 
 
 #### Runtime Types
@@ -362,8 +362,8 @@ TODO: Should we add `br_on_null` for completeness?
   - multiple invocations of this instruction yield the same observable RTTs
   - this is a *constant instruction*
 
-* `rtt.sub <n> <typeidx1> <typeidx2>` returns an RTT for `typeidx2` as a sub-RTT of a the parent RTT operand for `typeidx1`
-  - `rtt.sub n $t1 $t2 : [(rtt n $t1)] -> [(rtt (n+1) $t2)]`
+* `rtt.sub <typeidx>` returns an RTT for `typeidx` as a sub-RTT of a the parent RTT operand
+  - `rtt.sub $t : [(rtt n $t')] -> [(rtt (n+1) $t)]`
     - iff `t2 <: t1`
   - multiple invocations of this instruction with the same operand yield the same observable RTTs
   - this is a *constant instruction*
@@ -482,7 +482,7 @@ The opcode for heap types is encoded as an `s33`.
 | 0xfb21 | `i31.get_s` |  |
 | 0xfb22 | `i31.get_u` |  |
 | 0xfb30 | `rtt.canon $t` | `$t : typeidx` |
-| 0xfb31 | `rtt.sub n $t1 $t2` | `n : u32`, `$t1 : typeidx`, `$t2 : typeidx` |
+| 0xfb31 | `rtt.sub $t` | `$t : typeidx` |
 | 0xfb40 | `ref.test $t` | `$t : typeidx` |
 | 0xfb41 | `ref.cast $t` | `$t : typeidx` |
 | 0xfb42 | `br_on_cast $l $t` | `$l : labelidx`, `$t : typeidx` |
@@ -506,8 +506,6 @@ See [GC JS API document](MVP-JS.md) .
 ## Questions
 
 * Enable `i31` as a type definition.
-
-* Add `br_on_null`?
 
 * Should reference types be generalised to *unions*, e.g., of the form `(ref null? i31? data? func? extern? $t?)`? Perhaps even allowing multiple concrete types?
 
--- a/proposals/gc/MVP.md
+++ b/proposals/gc/MVP.md
@@ -22,7 +22,7 @@ All three proposals are prerequisites.
 
 #### Heap Types
 
-[Heap types](https://github.com/WebAssembly/function-references/blob/master/proposals/function-references/Overview.md#types) classify the target of a reference and are extended:
+[Heap types](https://github.com/WebAssembly/function-references/blob/master/proposals/function-references/Overview.md#types) classify reference types and are extended:
 
 * `any` is a new heap type
   - `heaptype ::= ... | any`
@@ -32,11 +32,19 @@ All three proposals are prerequisites.
   - `heaptype ::= ... | eq`
   - the common supertype of all referenceable types on which comparison (`ref.eq`) is allowed
 
+* `data` is a new heap type
+  - `heaptype ::= ... | data`
+  - the common supertype of all compound data types, like struct and array types and possibly host-defined types, for which casts are allowed
+
 * `i31` is a new heap type
   - `heaptype ::= ... | i31`
   - the type of unboxed scalars
 
 * Note: heap types `func` and `extern` already exist via [reference types proposal](https://github.com/WebAssembly/reference-types), and `(ref null? $t)` via [typed references](https://github.com/WebAssembly/function-references)
+
+We distinguish these *abstract* heap types from *concrete* heap types `(type $t)`.
+Each abstract heap type is a supertype of a class of concrete heap types.
+Moreover, they form a small [subtype hierarchy](#subtyping).
 
 
 #### Reference Types
@@ -49,15 +57,19 @@ New abbreviations are introduced for reference types in binary and text format, 
 * `eqref` is a new reference type
   - `eqref == (ref null eq)`
 
+* `dataref` is a new reference type
+  - `dataref == (ref data)`
+
 * `i31ref` is a new reference type
   - `i31ref == (ref i31)`
 
 
 #### Value Types
 
-* `rtt <n> <heaptype>` is a new value type that is a runtime representation of the static type `<heaptype>` and has `n` dynamic supertypes (see [Runtime types](#runtime-types))
-  - `valtype ::= ... | rtt <n> <heaptype>`
-  - `rtt n t ok` iff `t ok`
+* `rtt <n>? <typeidx>` is a new value type that is a runtime representation of the static type `<typeidx>`
+  - `valtype ::= ... | rtt <n>? <typeidx>`
+  - `rtt n? t ok` iff `t ok`
+  - the constant `n`, if present, encodes the static knowledge that this type has `n` dynamic supertypes (see [Runtime types](#runtime-types))
 
 
 #### Type Definitions
@@ -82,6 +94,8 @@ New abbreviations are introduced for reference types in binary and text format, 
   - `unpacked(t) = t`
   - `unpacked(pt) = i32`
 
+TODO: Need to be able to use `i31` as a type definition.
+
 
 #### Subtyping
 
@@ -94,21 +108,47 @@ In addition to the [existing rules](https://github.com/WebAssembly/function-refe
 * every type is a subtype of `any`
   - `t <: any`
 
+* `dataref` is a subtype of `eqref`
+  - `data <: eq`
+  - TODO: provide a way to make data types non-eq, especially immutable ones?
+
 * `i31ref` is a subtype of `eqref`
   - `i31 <: eq`
 
-* Any concrete type is a subtype of `eq` if its not a function
-  - `(type $t) <: eq`
+* Any concrete type is a subtype of either `data` or `func`
+  - `(type $t) <: data`
      - if `$t = <structtype>` or `$t = <arraytype>`
-     - or `$t = type rt` and `rt <: eq` (imports)
-  - TODO: provide a way to make data types non-eq, especially immutable ones
+     - or `$t = type ht` and `rt <: data` (imports)
+  - `(type $t) <: func`
+     - if `$t = <functype>`
+     - or `$t = type ht` and `rt <: func` (imports)
+
+Note: This creates a hierarchy of *abstract* Wasm heap types that looks as follows.
+```
+      any
+     /   \
+   eq    func
+  /  \
+i31  data
+```
+All *concrete* heap types (of the form `(type $t)`) are situated below either `data` or `func`.
+
+In addition, the abstract heap type `extern` is also a subtype of `any`.
+Its interpretation is defined by the host environment.
+It may contain additional host-defined types that are neither of the above three leaf type categories.
+It may also overlap with some or all of these categories, as would be observable by applying a classification instruction like `ref.is_func` to a value of type `externref`.
+The possible outcomes of such an operation hence depend on the host environment.
+(For example, in a JavaScript embedding, `externref` could be inhabited by all JS values -- which is a natural choice, because JavaScript is untyped; but some of its values are JS-side representations of Wasm values per the JS API, and those can also be observed as `data` or `func` references. Another possible interpretation could be that `data` is disjoint from `extern`, which would be determined by the coercions allowed by the JS API at the JS/Wasm boudary.)
+
+Note: In the future, this hierarchy could be refined to distinguish compound data types that are not subtypes of `eq`.
 
 
 #### Value Types
 
-* `rtt n t` is a subtype of `any`
-  - `rtt n t <: any`
-  - Note: `rtt n t1` is *not* a subtype of `rtt n t2`, even if `t1` is a subtype of `t2`; such subtyping would be unsound, since RTTs are used in both co- and contravariant roles (e.g., both when constructing and consuming a reference)
+* `rtt n $t` is a subtype of `rtt $t`
+  - `rtt n $t1 <: rtt $t2`
+    - if `$t1 == $t2`
+  - Note: `rtt n? $t1` is *not* a subtype of `rtt n? $t2`, if `$t1` is merely a subtype of `$t2`; such covariant subtyping would be unsound, since RTTs are used in both co- and contravariant roles (e.g., both when constructing and consuming a reference)
 
 
 ##### Defined Types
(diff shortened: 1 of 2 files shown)


proposals/gc/MVP.md
```diff
@@ -94,21 +108,47 @@ In addition to the [existing rules](https://github.com/WebAssembly/function-refe
...
+i31  data
+```
+All *concrete* heap types (of the form `(type $t)`) are situated below either `data` or `func`.
+
+In addition, the abstract heap type `extern` is also a subtype of `any`.
+Its interpretation is defined by the host environment.
+It may contain additional host-defined types that are neither of the above three leaf type categories.
+It may also overlap with some or all of these categories, as would be observable by applying a classification instruction like `ref.is_func` to a value of type `externref`.
+The possible outcomes of such an operation hence depend on the host environment.
+(For example, in a JavaScript embedding, `externref` could be inhabited by all JS values -- which is a natural choice, because JavaScript is untyped; but some of its values are JS-side representations of Wasm values per the JS API, and those can also be observed as `data` or `func` references. Another possible interpretation could be that `data` is disjoint from `extern`, which would be determined by the coercions allowed by the JS API at the JS/Wasm boudary.)
```
RossTate, line 141: This would cause problems for code migration. That is, I imagine we would like it to be easy for a program component written in JS to switch to GC-Wasm (possibly with JS-decorated `rtt`s for JS interop). But the suggestion here would make it so that the switch would cause previously accepted coercions (in other program components) to `externref` to become invalid. This means a converted module would have to create JS proxies (rather than `rtt` decorations) around all of its escaping references in order to ensure backwards compatibility.

dcodeIO: You had my attention at

> we would like it to be easy for a program component written in JS to switch to GC-Wasm

and I'd like to note that this isn't exactly what we are gravitating towards currently for other reasons, in part because of your earlier comments elsewhere. Rather seems like we are doing a slingshot around Jupiter while still hoping that there'll remain a faint signal of code migration in deep space, so we can point at it and say "look, it's still there". 🪐 🚀 💥

RossTate: @dcodeIO I have implemented systems with a typed language (similar to GC-Wasm) and an untyped OO language (similar to JS) that interop efficiently and in a manner that supports code migration. One reason I posed [this extension](https://github.com/WebAssembly/call-tags/blob/master/proposals/call-tags/Overview.md#extension) to the Call Tags proposal is that we found this mechanism sped up method calls in untyped code when invoked on objects originating from typed code. I have not, however, written up how to add such efficient interop to GC-Wasm and imports/exports because some of the techniques are not compatible with the current MVP (or Post-MVP). So I share the goal you have in mind, but I am trying to achieve it by first getting the right foundations in place for supporting efficient interop.

dcodeIO: I see, thanks for doing that! Whenever you have something of interest to me in this domain, feel free to ping me :)

proposals/gc/MVP.md
```diff
@@ -266,62 +293,112 @@ Tentatively, support a type of guaranteed unboxed scalars.
...
 
 #### Casts
 
-* `ref.test <heaptype1> <heaptype2>` tests whether a reference value's [runtime type](#values) is a [runtime subtype](#runtime) of a given RTT
-  - `ref.test t1 t2 : [(ref null t1) (rtt n t2)] -> [i32]`
-     - iff `t2 <: t1`
+RTT-based casts can only be performed with respect to concrete types, and require a data or function reference as input, which are known to carry an RTT.
+
+* `ref.test <typeidx>` tests whether a reference value's [runtime type](#values) is a [runtime subtype](#runtime) of a given RTT
+  - `ref.test $t : [(ref null ht) (rtt n $t)] -> [i32]`
```
titzer, line 372: I think this should be `(rtt n? $t)`, i.e. the depth `n` is optional.

proposals/gc/MVP.md
```diff
@@ -266,62 +293,112 @@ Tentatively, support a type of guaranteed unboxed scalars.
...
+Note: The condition `(type $t) <: ht` isn't needed for soundness of any of the above instructions. If false, the check merely is statically known to fail. Should it be removed?
+
+
+#### Constant Expressions
+
+In order to allow RTTs to be initialised as globals, the following extensions are made to the definition of *constant expressions*:
+
+* `rtt.canon` is a constant instruction
+* `rtt.sub` is a constant instruction
+* `global.get` is a constant instruction and can access preceding global definitions, not just imports as in the MVP
```
titzer, line 400: Preceding immutable global definitions.

proposals/gc/MVP.md
```diff
@@ -173,14 +215,13 @@ This can compile to machine code that (1) reads the RTT from `$x`, (2) checks th
...
-* Creating a structure or array optionally allows supplying a suitable RTT value to represent its runtime type; it is `any` if none is given.
+* Creating a structure or array requires supplying a suitable RTT value to represent its runtime type.
 
-* Each reference value has an associated runtime type:
-  - For structures or arrays, it is the RTT value provided upon creation, or `anyref` if none.
-  - For functions, it is the RTT value for the function's type.
-  - For `i31ref` references it is the RTT value for `i31ref`.
+* Reference values of data or function type have an associated runtime type:
+  - for structures or arrays, it is the RTT value provided upon creation,
+  - for functions, it is the RTT value for the function's type.
```
titzer, line 222: Does this mean that functions from the function references proposal would implicitly have `(rtt.canon $t)`? How would one specify another RTT for a function?

proposals/gc/MVP.md
```diff
@@ -266,62 +293,112 @@ Tentatively, support a type of guaranteed unboxed scalars.
...
+* `ref.test <typeidx>` tests whether a reference value's [runtime type](#values) is a [runtime subtype](#runtime) of a given RTT
+  - `ref.test $t : [(ref null ht) (rtt n $t)] -> [i32]`
+    - iff `ht <: data` or `ht <: func`
+    - and `(type $t) <: ht`
   - returns 1 if the first operand is not null and its runtime type is a sub-RTT of the RTT operand, 0 otherwise
 
-* `ref.cast <heaptype1> <heaptype2>` casts a reference value down to a type given by a RTT representation
-  - `ref.cast t1 t2 : [(ref null t1) (rtt n t2)] -> [(ref t2)]`
-     - iff `t2 <: t1`
+* `ref.cast <typeidx>` casts a reference value down to a type given by a RTT representation
```
titzer, line 377: What about adding the (optional) depth of the RTT as an immediate here? Since there are now two kinds of RTT checks, one where the depth is known and one where the depths are unknown, which kind of check it is depends on the type of the incoming RTT value. An interpreter doesn't know that statically, so it needs to inspect the RTT value to know whether it is a "depth carrying" RTT or a "depth independent" RTT.

proposals/gc/MVP.md
```diff
@@ -132,33 +172,35 @@ In addition to the [existing rules](https://github.com/WebAssembly/function-refe
...
 Example: Consider three types and corresponding RTTs:
 ```
 (type $A (struct))
 (type $B (struct (field i32)))
 (type $C (struct (field i32 i64)))
 
-(global $rttA (rtt 1 $A) (rtt.sub $A (rtt.canon any)))
-(global $rttB (rtt 2 $B) (rtt.sub $B (global.get $rttA)))
-(global $rttC (rtt 3 $C) (rtt.sub $C (global.get $rttB)))
+(global $rttA (rtt 0 $A) (rtt.canon $A))
```
titzer, line 201: Is it possible to create a depth-0 RTT without using `rtt.canon`? Otherwise, e.g. in a class-based language, two disjoint roots of different class hierarchies that just happen to be empty will get lumped together by using `rtt.canon`. They'd have to declare a superfluous empty struct to use as their parent to get different hierarchies started.

proposals/gc/MVP.md
rossberg: Yes, for backwards compatibility, functions have a canonical RTT. There was a brief discussion somewhere about a mechanism for picking custom RTTs for functions, but I can't find it anymore. Added a short section to Post-MVP.md. Do you see an immediate use case for it?

proposals/gc/MVP.md
rossberg: Yup, fixed here and below.

proposals/gc/MVP.md
rossberg: Yeah, this is part of a wider discussion. In the light of recent discussions, it seems like we rather might want to remove the immediate altogether in cases like this. So I suggest leaving as is until we have resolved that discussion.

For an interpreter, I would assume that it simply ignores the static depth altogether and just reads it off the RTT value uniformly. The win of using the static value may already be tiny for a compiler, I doubt that it matters in an interpreter.

proposals/gc/MVP.md
rossberg: Yeah, changed for conservativeness's sake. Though I wonder if there is a hard reason for restricting it, given that no global can actually be mutated yet in that phase?

proposals/gc/MVP.md
titzer: I am more worried of needing to eagerly compute `rtt.canon` for a function type in `func.ref`, because keeping around enough metadata to lazily compute it might be even worse.

proposals/gc/MVP.md
titzer: I think you would get different results because isn't an RTT with depth a subtype of RTT without? And thus the static type and dynamic type of the RTT may be different, so an interpreter could not know the difference by inspecting the value alone.

proposals/gc/MVP.md
rossberg: Hm, at least uses of `ref.func` need to be pre-declared, so you don't have to that for functions only used second-class. But for the others, what alternative would you suggest? I don't know how we could avoid that first-class functions carry their type information somehow, if we want to allow casts on them.

proposals/gc/MVP.md
rossberg: Yes, `(rtt n $t) <: (rtt $t)`, see the definition of subtyping above. This is so that you can abstract from the depth where you don't want to expose or depend on it. It can't result in a different value, though. It's just the length of the supertype vector in the value, so presumably needs to be derivable from the value representation anyway.

proposals/gc/MVP.md
titzer: I know, I don't have a good answer right now, other than a vague feeling that `func.ref` and `func.bind` are analogous to allocations that would need explicit RTT values.

proposals/gc/MVP.md
titzer: I agree. What I mean is that a cast that uses an RTT of static type `(rtt n $t)` checks only that the RTT at index `n` in the supertype vector, whereas a cast with RTT of static type `(rtt $t)` checks all entries in the supertype vector. Because of subtyping on RTT values, an RTT value allocated as `(rtt n $t)` is a subtype of `(rtt $t)` and can end up at cast site of static type `(rtt $t)`, so it is not possible to distinguish the two cases based on the RTT value's dynamic type alone.

rossberg: > whereas a cast with RTT of static type (rtt $t) checks all entries in the supertype vector.

Hm, that's not the intention. Why should it have to do that? The only difference should be how it determines the slot index, everything else should be the same.

proposals/gc/MVP.md
RossTate: I am referring to the (existing) coercions to `externref` in the JS API.

rossberg: You'd have to be a little more concrete. At the boundary, externref does not imply any observable coercion, nor would its semantics change.

RossTate: Kotlin currently compiles to JS, representing Kotlin objects as JS objects. That means those Kotlin-as-JS objects can currently be given to wasm programs as `externref` values. Now if a Kotlin program is compiled to GC-wasm instead (maybe with some JS decoration in the `rtt` for JS interop), then your suggestion above would mean that Kotlin-as-wasm objects would not be passable to wasm programs as `externref` values. Thus previously working interactions would break by switching to wasm even if one does the job of decorating the rtts of their wasm references with JS info for JS interop.

rossberg: If a module explicitly asks to import an externref then switching between host and Wasm won't generally work. But I don't know what's surprising about that, you'd get what you asked for. Of course, there is no point in ever doing that, anyref is what you want on module boundaries.

RossTate: This is a surprising response. You're saying that the following is fine with you. Say you have a web program P using a third-party library L both currently implemented in JS. Suppose P changes to use wasm but in a backwards-compatible manner, using `externref` to handle JS values (such as L's) so that wasmP + L behaves as before. Suppose L changes to use wasm but in a backwards-compatible manner, using `externref` to handle JS values (such as P's) so that P + wasmL behaves as before. Then wasmP + wasmL (with JS glue) will not work because their uses of `externref` can no longer denote the other's values. That seems to make wasm interop with the web ecosystem rather fragile.

brion: I would assume both wasmL and wasmP would be using `externref` at the boundary, as they expect to be compatible with JavaScript on the other end. This would seem to "just work" if I understand correctly?

brion: Ah, I see now -- if I import a function from another module I have to specify the arguments' types on the import. If I specify `externref`, I can't call the other module's function with a GC value if it's now impossible to cast to `externref`.

A workaround could be to import a host function for each type that returns its argument as an `externref`. This might be awkward and slow compared to casts...

proposals/gc/MVP.md
titzer: Ok, well I misunderstood then. I was under the impression that `rtt.sub` was generative--i.e. it would generate a new RTT value each time. If RTT's are also structural, they are effectively useless for encoding a nominal class hierarchy. That's pretty much a deal breaker unless I am missing something else.

rossberg: Thus the TODO. Though let me point out the obvious: even with generative RTTs added, Wasm-level RTTs and casts will never be sufficient to directly implement the majority of source-language nominal hierarchies and casts -- for the same reason that Wasm's static type can never encode all source type systems. You can piggy-back source casts on Wasm casts in some limited cases, but don't expect that to be possible in every case. In the general case, you have to implement your own tagging system in user space. (And that makes me wonder whether generative RTTs are even worth adding to the MVP.)

proposals/gc/MVP.md
rossberg: > Suppose L changes to use wasm but in a backwards-compatible manner, using `externref` to handle JS values

L would not be able to provide Wasm data values as externref under that alternative semantics. This would not be a possible change, let alone a backwards-compatible one. If you truly said externref you should mean it. You are clearly leaking your implementation choice if you do.

Again, externref is not the type you ever want to use as an abstract type at the boundary -- if you leak implementation details, you are artificially restricting your evolution paths, and get what you ask for. You always want to use type imports or anyref, which properly abstract away a module's implementation choice.

proposals/gc/MVP.md
titzer: I'd argue that generative RTT values are actually cheaper and simpler to implement in an engine than structural RTT values, and I can verify that they are enough to implement the ubiquitous class casts in Java such that a wasm-level cast implies a Java-level cast, even, and perhaps especially, with classloaders. (They are not sufficient for interface or array casts, however). Without generative RTTs then a Java implementation is forced into doing both a wasm-level cast and a Java-level check, because with structural RTTs there is no way to make a wasm-level cast imply a Java-level check.

rossberg: Agreed, but let's keep generative RTTs for a different PR. This one does not change that aspect, so should be orthogonal.

proposals/gc/MVP.md
RossTate: According to the current JS API for the MVP, a wasm module can export wasm references that have well-defined interactions in JS code. So wasmL can use wasm references where L used JS references, and if done right the change will be backwards compatible with other JS code, such as P. So wasmL imports `externref` values to reference (foreign) JS values, but exports reference values that the JS API "converts" (without wrapping) to JS values.

> Again, externref is not the type you ever want to use as an abstract type at the boundary

The reference types proposal says that "Any JS value can be passed as externref to a Wasm function, stored in a global, or in a table". So people are generating programs with the expectation that `externref` represents any JS value. Changing that to "Another possible interpretation could be that `data` is disjoint from `extern`" would make it difficult for new wasm programs to interact with old wasm programs in the web ecosystem that were not written with this distinction in mind.

brion: > You always want to use type imports or anyref, which properly abstract away a module's implementation choice.

* If I should always use `anyref`, why does `externref` exist?
* How do you reference JavaScript duck typing behavior in a type import?

brion: Also to clarify; if a JS value containing a Wasm GC struct or array reference is passed to a Wasm function that defines an argument as type `externref`, what happens? Does it trap now?

brion: Note that it makes a lot of sense to use `externref` for JS objects, since you might have to call methods and access properties on them and this is a JS-specific interface that requires importing helper functions like `Reflect.get`, `Reflect.construct`, and `Reflect.apply` from JavaScript, which will accept JavaScript values and so be imported using `externref` as their arguments and return values. There's little benefit I can imagine to choosing `anyref` over `externref` there.

If your API doesn't involve properties or prototype-bound methods on the JS objects, but only involves passing the objects as arguments to functions which you can import in Wasm, then maybe you wouldn't want to make it `externref`.

Or am I missing something and there's some other way to access regular JavaScript objects with properties and prototype-bound methods that doesn't involve calling out to JavaScript from Wasm?

rossberg: @brion:
> * If I should always use anyref, why does externref exist?
Externref is a wart that only came to be because we removed subtyping from the reference-types proposal at the last minute and thus anyref had to go.

> * How do you reference JavaScript duck typing behavior in a type import?
Sorry, I don't understand the question. There is no duck typing in Wasm.

> if a JS value containing a Wasm GC struct or array reference is passed to a Wasm function that defines an argument as type externref, what happens? Does it trap now?
The most likely interpretation would be that it is "coerced" in some form.

> Note that it makes a lot of sense to use externref for JS objects, since you might have to call methods and access properties on them and this is a JS-specific interface that requires importing helper functions like Reflect.get, Reflect.construct, and Reflect.apply from JavaScript, which will accept JavaScript values and so be imported using externref as their arguments and return values. There's little benefit I can imagine to choosing anyref over externref there.
All this works just fine with anyref. However, if you really want to explicitly use JS objects in such an interface, then you are indeed not interested in abstraction or the ability to transition them to Wasm objects instead. Then externref seems fine, and datarefs being disjoint would not be relevant.

@RossTate:
> According to the current JS API for the MVP
If you mean the GC MVP, then under a hypothetical design where dataref and externref were separate, it obviously would have to be adjusted accordingly.

Folks, before we waste more energy on this: this parenthesis was meant as an illustrative remark. I am not actually proposing to do it that way. It merely illustrates the design space that exists for hosts to define the meaning of externref. I realise that JS is not the greatest example to use, but it is the one that everybody knows. I added a clarifying sentence to point that out and (hopefully) avoid possible confusion.

brion: > All this works just fine with anyref. However, if you really want to explicitly use JS objects in such an interface, then you are indeed not interested in abstraction or the ability to transition them to Wasm objects instead. Then externref seems fine, and datarefs being disjoint would not be relevant.

The scenario is that we have two existing JavaScript modules which both provide JavaScritp APIs that interact with each other.

So yes, the APIs need to remain JavaScript-compatible when implementations are changed from JavaScript to Wasm.

That is why there are questions about how to implement that kind of code migration behavior, since "code migration" is a significant goal.

brion: I should clarify that the APIs already exist in this scenario! They were not designed with anything Wasm in mind; they are JS code that already exists.

rossberg:
@titzer, @jakobkummerow, any further thoughts on this change?

rossberg:
@titzer, @jakobkummerow, ping.

jakobkummerow (approved):
I'm fine with these changes. I have a few comments. (Also, sorry for the delay.)

I'd like to point out that approving these changes to RTTs does not mean that I would be opposed to simplifying things even more by dropping RTTs entirely and folding them into the static types. That discussion is still ongoing, elsewhere.

Note to observers: approving these changes does not constitute a commitment to any timeline for updating the V8-based prototype accordingly. If you have an opinion on that (e.g., you're eagerly awaiting these changes, or on the contrary, you'd prefer if the prototype's behavior didn't change for a while), please let me know.

proposals/gc/MVP.md
```diff
@@ -273,62 +300,112 @@ Tentatively, support a type of guaranteed unboxed scalars.
...
 
 * `ref.as_i31` converts to an integer reference
   - `ref.as_i31 : [anyref] -> [i31ref]`
   - traps if reference is not an integer
-  - equivalent to `(rtt.canon i31) (ref.cast)`
+  - equivalent to `(block $l (param anyref) (result i31ref) (br_on_i31 $l) (unreachable))`
+
+Note: There are no instructions to check for `externref`, since that can consist of a diverse set of different object representations that would be costly to check for exhaustively.
+
+TODO: Should we add `br_on_null` for completeness?
```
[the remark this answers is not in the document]
rossberg: Ah, right. :) Added comment with links for ref.is_null and br_on_null.

proposals/gc/MVP.md
```diff
@@ -273,62 +300,112 @@ Tentatively, support a type of guaranteed unboxed scalars.
...
-  - `n = 0` iff `t = any`, and `n = 1` otherwise
+* `rtt.canon <typeidx>` returns the RTT of the specified type
+  - `rtt.canon $t : [] -> [(rtt 0 $t)]`
   - multiple invocations of this instruction yield the same observable RTTs
   - this is a *constant instruction*
-  - equivalent to `(rtt.sub 1 any t (rtt.canon any))`, except when `t` itself is `any`
 
-* `rtt.sub <n> <heaptype1> <heaptype2>` returns an RTT for `heaptype2` as a sub-RTT of a the parent RTT operand for `heaptype1`
-  - `rtt.sub n t1 t2 : [(rtt n t1)] -> [(rtt (n+1) t2)]`
+* `rtt.sub <n> <typeidx1> <typeidx2>` returns an RTT for `typeidx2` as a sub-RTT of a the parent RTT operand for `typeidx1`
```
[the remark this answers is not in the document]
rossberg: Fair enough, I included that change. Especially since the example already assumes it. :)

proposals/gc/MVP.md
```diff
@@ -418,8 +505,12 @@ See [GC JS API document](MVP-JS.md) .
 
 ## Questions
 
-* Should RTT presence be made explicit in struct types and ref types?
-  - for example, `(struct rtt ...)` and `rttref <: anyref`
-  - only these types would be castable
+* Enable `i31` as a type definition.
+
+* Add `br_on_null`?
```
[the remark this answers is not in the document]
rossberg: Removed.

rossberg (commented):
Thanks for the comments!

proposals/gc/MVP.md
titzer: Ok.

proposals/gc/MVP.md
```diff
@@ -180,14 +222,13 @@ This can compile to machine code that (1) reads the RTT from `$x`, (2) checks th
...
-* Each reference value has an associated runtime type:
-  - For structures or arrays, it is the RTT value provided upon creation, or `anyref` if none.
-  - For functions, it is the RTT value for the function's type.
-  - For `i31ref` references it is the RTT value for `i31ref`.
+* Reference values of data or function type have an associated runtime type:
+  - for structures or arrays, it is the RTT value provided upon creation,
+  - for functions, it is the RTT value for the function's type.
 
-* The so-defined runtime type is the only type information that can be discovered about a reference value at runtime; a structure or array with RTT `any` thereby is fully opaque to runtime type checks (and an implementation may choose to optimize away its RTT).
+* Note: as a future extension, we could allow a value's RTT to be a supertype of the value's actual type. For example, a structure or array with RTT `any` would become fully opaque to runtime type checks, and an implementation may choose to optimize away its RTT.
```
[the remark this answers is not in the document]
titzer: I am confused by the comment as written. I thought the point of this PR was to eliminate the possibility of having an RTT for any, but you are suggesting adding it back in a future extension?

proposals/gc/MVP.md
titzer: The difference is that using the types (of RTTs) inferred from validation, a JIT would use the slot index inferred from the RTT's static type and emit a check _only_ of that slot index, and not a search, whereas an interpreter would not have this information from the inferred RTT type, as it only has the RTT values.

Thus I think it's necessary to have the optional depth here, otherwise you force JITs to emit a search too.

proposals/gc/MVP.md
```diff
@@ -139,33 +179,35 @@ In addition to the [existing rules](https://github.com/WebAssembly/function-refe
 
 #### Runtime Types
 
-* Runtime types (RTTs) are explicit values representing types at runtime; a value of type `rtt <n> <heaptype>` is a dynamic representative of static type `<heaptype>`.
+* Runtime types (RTTs) are explicit values representing concrete types at runtime; a value of type `rtt <n>? <typeidx>` is a dynamic representative of the static type `<typeidx>`.
```
jakobkummerow, line 182: This needs an update to the binary encoding section. The decoder needs a way to know whether the `<n>` is present or not. One possible solution is to have different encodings for rtt-with-n and rtt-without-n. Another possible solution is to use a sentinel value for absent `<n>`, such as -1. I'm sure there are other possible solutions too.
(The binary encoding section should also be updated to reflect the heaptype->typeidx change.)

proposals/gc/MVP.md
```diff
@@ -273,62 +300,112 @@ Tentatively, support a type of guaranteed unboxed scalars.
...
 
 * `i31.get_u` extracts the value, zero-extending
   - `i31.get_u : [i31ref] -> [i32]`
 
 * `i31.get_s` extracts the value, sign-extending
   - `i31.get_s : [i31ref] -> [i32]`
 
-Perhaps also the following short-hands:
+
+#### Classification
```
jakobkummerow, line 312: Is it intentional that there are no instructions for eqref, i.e. `ref.is_eq`/`ref.as_eq`/`br_on_eq`? Previously it was possible to cast from anyref ot eqref, though I'm not sure how relevant that is in practice. In particular, if we end up making i31 an attribute on references (like nullability), then `ref.is_data` will effectively subsume `ref.is_eq`.

jakobkummerow (commented):
Two more comments.

proposals/gc/MVP.md
rossberg: Ah, indeed. I added opcode `-0x18` for rtt without n.

proposals/gc/MVP.md
rossberg: Yes. Eqref isn't really a basic category of values but rather a constraint on anyref. I can't think of a good use case to cast to it, and it doesn't seem like that would be a simple operation to implement either.

proposals/gc/MVP.md
```diff
@@ -432,6 +432,7 @@ This extends the [encodings](https://github.com/WebAssembly/function-references/
 | -0x15  | `(ref ht)`      | `ht : heaptype (s33)` | from funcref proposal |
 | -0x16  | `i31ref`        |            | |
 | -0x17  | `(rtt n ht)`    | `n : u32`, `ht : heaptype (s33)` | |
+| -0x18  | `(rtt ht)`      | `ht : heaptype (s33)` | |
```
jakobkummerow, line 435: Here and in the line above: s/heaptype (s33)/typeidx (u32)/

rossberg: Oops, fixed.

proposals/gc/MVP.md
```diff
@@ -354,7 +431,8 @@ This extends the [encodings](https://github.com/WebAssembly/function-references/
 | -0x14  | `(ref null ht)` | `ht : heaptype (s33)` | from funcref proposal |
 | -0x15  | `(ref ht)`      | `ht : heaptype (s33)` | from funcref proposal |
 | -0x16  | `i31ref`        |            | |
-| -0x17  | `(rtt n ht)`    | `n : u32`, `ht : heaptype (s33)` | |
+| -0x17  | `(rtt n $t)`    | `n : u32`, `$t : typeidx` | |
```
jakobkummerow, line 434: Since you're here: the "Heap Types" section right below has a typo in the `i31` row (line 448), where it currently says -0x17. It should be consistent with line 433 and say -0x16.

rossberg: Thanks, fixed.

Issue #152, opened 2020-10-12
Casting null


titzer:
Currently, `ref.cast` is spec'd to trap on null, and always return a non-null reference. For some languages, e.g. Java, casts of a null succeed. Thus to implement this behavior, user code must insert a branch checking for null first. To avoid making that check for null redundant with the null check in `ref.cast`,  `br_on_null` is a natural fit, since it refines the type to be non-null in the fallthrough. But `br_on_null` pops the null value off of the operand stack, so we end up with a sequence like:

```
(block $l1 (ref $t2))
  (block $l2)
    (br_on_null $l2 x)
    (ref_cast $t1 $t2)  // <-- benefits from knowing x non-null
    (br $l1)
  (end)
  (ref_null $t2)
(end)
```

Or another option:

```
(ref_is_null x)
(if (ref $t2))
  (ref_null $t2)
 (else)
  (ref_cast $t1 $t2 x) // <-- x not known to be non-null here
(end)
```

(note the above are simplified, omitting the RTT values)

Instead both of these sequences could be simplified if we change the semantics of `ref_cast`. For example, if `ref_cast` has a type immediate that is allowed to be a nullable type, then the nullability could be used to indicate whether `null` is allowed to succeed, or should trap.

skuzmich:
Bumped into this in Kotlin compiler. It would be real nice to have a version of `ref_cast` that doesn't trap on null.  

Also, in second example above, value of `x` is used twice, requiring additional `local.tee` and `local.get`.

Horcrux7:
I does not see any sense that I can't cast a NULL value to another nullable value. This is also not consistence because a null value required an type for `ref.null`.

titzer:
The use case is, e.g. downcasting from (the wasm representation of) `java/lang/Object` to `java/lang/String`. In Java, this cast will succeed if the input value is `null`.

rossberg:
Hm, I see the reasoning, but there is a trade-off. It would change the typing from
```
ref.cast : [(ref null t1) (rtt $t2)] -> [(ref $t2)]
```
to
```
ref.cast : [(ref nl t1) (rtt $t2)] -> [(ref nl $t2)]  (where nl in {null, epsilon})
```
This would be a potential extra cost in contexts where the producer knows out of band that the argument can't be null. It would be forced to insert an additional `ref_as_non_null`, i.e., an extra null check, in some cases. But that's probably fine and easy to optimize in the engine in most cases? Or is it worth having two versions of the instruction?

[...]

PR #161, opened 2020-11-19
Make ref.cast succeed on null

rossberg:
This changes `ref.cast` to propagate null instead of trapping. Addresses #152.

(For now, not introducing two variants of the instruction -- let's see how common the other case really is to justify duplication.)


titzer (approved):
lgtm, thanks!

PR #164, opened 2020-12-02
Fix a typo in the overview

### LINKED ISSUE #163: Can a field be an embedded value?
The [type grammer](https://github.com/WebAssembly/gc/blob/master/proposals/gc/Overview.md#type-grammar) says a field of an array or a struct is a storage type, which is a value type (ignoring packed types), which is either a number like i32, or a reference type.

However, the examples on the same page have this:
```wat
(type $vector (array (mut f64)))
(type $matrix (array (type $vector)))
```

Is a matrix an array of *references* to vectors, or an array of vector *values*?

Is "type" a typo for "ref" perhaps in that example? Or is it an alias for "ref"? Or do we actually intend to allow embedded values like e.g. Go does?

(If we allow embedded values, then the Binaryen type system for fields is insufficient, as it cannot represent them, which is why I ran into this. cc @tlively @dcodeio)

kripken:
Fixes #163

### DIFF
--- a/proposals/gc/Overview.md
+++ b/proposals/gc/Overview.md
@@ -641,7 +641,7 @@ def_type       ::=  <data_type> | <func_type>
 ```
 where `value_type` is the type usable for parameters, local variables and the operand stack, and `def_type` describes the types that can be defined in the type section.
 
-Note that an additional restruction on the above grammar is that array fields must be mutable.
+Note that for the MVP, an additional restriction on the above grammar is that array fields must be mutable.
 
 ### Type Recursion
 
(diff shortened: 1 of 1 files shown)


proposals/gc/Overview.md
```diff
@@ -641,6 +641,7 @@ def_type       ::=  <data_type> | <func_type>
 ```
 where `value_type` is the type usable for parameters, local variables and the operand stack, and `def_type` describes the types that can be defined in the type section.
 
+Note that an additional restruction on the above grammar is that array fields must be mutable.
```
rossberg, line 644:
```suggestion
Note that for the MVP, an additional restriction on the above grammar is that array fields must be mutable.
```

rossberg (approved):
Ah, thanks!

PR #177, opened 2021-01-13
Fix def of sub-RTT

### LINKED ISSUE #170: Understanding RTT semantics
Probably just me (not having a CS background) but personally I've found this a little hard to follow. Perhaps some more examples or reftests might help - or are those present somewhere I didn't find?

One specific question is what RTT casts are valid. For example,
```wat
x = (rtt.canon A)
y = (rtt.sub B x)
z = (rtt.sub C y)

m = (rtt.canon A)
n = (rtt.sub C m)
```
Here `z -> y -> x`, and `n -> m`, with similar types except the second chain "skips" the type B in the middle. Is it valid to cast an object with RTT `z` using an RTT `n`? The object has the right wasm type (C in both cases), but different RTT "chains". So it's memory safe, but perhaps not valid to cast?

rossberg:
Fix #170.

### DIFF
--- a/proposals/gc/MVP.md
+++ b/proposals/gc/MVP.md
@@ -185,8 +185,12 @@ Note: In the future, this hierarchy could be refined to distinguish compound dat
 
 * There is a runtime subtyping hierarchy on RTTs; creating an RTT allows providing a *parent type* in the form of an existing RTT.
 
+* An RTT value r1 is *equal* to another RTT value r2 iff they both represent the same static type and either of the following holds:
+  - r1 and r2 both have no parents, or
+  - r1 and r2 both have equal RTT values as parents.
+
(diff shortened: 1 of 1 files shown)


conrad-watt (approved):
If `rtt.fresh` is introduced, this definition will need to be tweaked. For now, this is the smallest change that gets us to the semantics described by the implementation note/@jakobkummerow's expectation.

This isn't my preferred semantics, but given that it seems to have been informally expected, I'll follow up separately.

rossberg:
Yes, rtt.fresh would obviously require extending the definition. (Though if we add static nominal types anyway, then rtt.fresh becomes moot and would be subsumed by rtt.canon on a nominal type, in which the definition remains adequate.)

Issue #183, opened 2021-01-20
Comments/Questions regarding classification opcodes


manoskouk:
Hello,
I have been implementing the latest version of the gc proposal in V8 and have come up with the following comments/questions:

- While `ref.cast` succeeds on null, it is not clear what should happen with the `ref.as_*` instructions. As stated, it seems they should all fail for null, but then `ref.as_func` should return `(ref func)` over `funcref == (ref null func)`. What is the intended behavior? 
- Similarly, `br_on_func` should branch with a `(ref func)` type.
- The typing of `br_on_* : [anyref]->[anyref]` instructions needlessly loosens the type of its argument. E.g., if we pass an `eqref` to `br_on_data` and the branch is not taken, we will end up with an `anyref` value instead of `eqref`. In my opinion, those instructions should return the type of their argument.
- The `br_on_*` instructions only work if the targeted label expects as result exactly the value produced by the branch instruction. This could be loosened to allow the branch type to have more values, or even no values (in which case the matched value is discarded). If both are allowed, the behavior of `(br_on_t $l arg)` where `arg: ta`. would be equivalent to `(if (result ta) (ref.is_t arg) (then (ref_as_t arg) (br $l)) (else arg))`.

rossberg:
Excellent suggestions! PTAL at #184 for a fix.

> * While `ref.cast` succeeds on null, it is not clear what should happen with the `ref.as_*` instructions. As stated, it seems they should all fail for null, but then `ref.as_func` should return `(ref func)` over `funcref == (ref null func)`. What is the intended behavior?
> * Similarly, `br_on_func` should branch with a `(ref func)` type.

Absolutely! Also refined the other `ref_as_*` and `br_on_*` typings accordingly.

> * The typing of `br_on_* : [anyref]->[anyref]` instructions needlessly loosens the type of its argument. E.g., if we pass an `eqref` to `br_on_data` and the branch is not taken, we will end up with an `anyref` value instead of `eqref`. In my opinion, those instructions should return the type of their argument.

Agreed, changed as well.

> * The `br_on_*` instructions only work if the targeted label expects as result exactly the value produced by the branch instruction. This could be loosened to allow the branch type to have more values, or even no values (in which case the matched value is discarded). If both are allowed, the behavior of `(br_on_t $l arg)` where `arg: ta`. would be equivalent to `(if (result ta) (ref.is_t arg) (then (ref_as_t arg) (br $l)) (else arg))`.

[...]

Issue #180, opened 2021-01-15
Missing binary encoding for dataref


manoskouk:
The encoding of the `dataref` value type and `data` heap type are missing from the binary format section.

rossberg:
Fixed via #184.

PR #184, opened 2021-01-21
Relax typing of classification instructions; opcode for dataref

rossberg:
Addresses #180 and #183.

### DIFF
--- a/proposals/gc/MVP.md
+++ b/proposals/gc/MVP.md
@@ -325,38 +325,43 @@ Tentatively, support a type of guaranteed unboxed scalars.
   - `ref.is_i31 : [anyref] -> [i32]`
 
 * `br_on_func <labelidx>` branches if a reference is a function
-  - `br_on_func $l : [anyref] -> [anyref]`
-    - iff `$l : [funcref]`
+  - `br_on_func $l : [(ref null ht)] -> [(ref null ht)]`
+    - iff `$l : [(ref null? ht')]`
+    - and `func <: ht'`
   - passes operand along with branch as a function
 
 * `br_on_data <labelidx>` branches if a reference is compound data
-  - `br_on_data $l : [anyref] -> [anyref]`
-    - iff `$l : [dataref]`
-  - passes operand along with branch as a function
+  - `br_on_data $l : [(ref null ht)] -> [(ref null ht)]`
+    - iff `$l : [(ref null? ht')]`
+    - and `data <: ht'`
+  - passes operand along with branch as data
 
 * `br_on_i31 <labelidx>` branches if a reference is an integer
-  - `br_on_func $l : [anyref] -> [anyref]`
-    - iff `$l : [i31ref]`
-  - passes operand along with branch as a function
+  - `br_on_i31 $l : [(ref null ht)] -> [(ref null ht)]`
+    - iff `$l : [(ref null? ht')]`
+    - and `i31 <: ht'`
+  - passes operand along with branch as a scalar
 
 * `ref.as_func` converts to a function reference
-  - `ref.as_func : [anyref] -> [funcref]`
+  - `ref.as_func : [anyref] -> [(ref func)]`
   - traps if reference is not a function
   - equivalent to `(block $l (param anyref) (result funcref) (br_on_func $l) (unreachable))`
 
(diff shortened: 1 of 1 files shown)


rossberg:
@manoskouk, PTAL.

proposals/gc/MVP.md
```diff
@@ -325,39 +325,44 @@ Tentatively, support a type of guaranteed unboxed scalars.
   - `ref.is_i31 : [anyref] -> [i32]`
 
 * `br_on_func <labelidx>` branches if a reference is a function
-  - `br_on_func $l : [anyref] -> [anyref]`
-    - iff `$l : [funcref]`
+  - `br_on_func $l : [(ref null ht)] -> [(ref null ht)]`
```
manoskouk, line 328: I think the typing of all `br_on_*` instructions, including `br_on_cast`, could be further refined to preserve their argument's nullability, similarly to `ref.cast`.

manoskouk (commented):
LGTM with one comment.

proposals/gc/MVP.md
rossberg: Good point, PTAL!

manoskouk (approved):
LGTM, thanks!

rossberg:
Ah, yes. I included that change, and simplified their typing rules. While at it, I also removed the explicit type index on those instructions, which is no longer consistent with the rest of the language. PTYAL.

manoskouk:
Looks good, thanks!

proposals/gc/MVP.md
```diff
@@ -379,29 +387,27 @@ TODO: Add the ability to generate new (non-canonical) RTT values to implement ca
...
-* `br_on_cast <labelidx> <typeidx>` branches if a value can be cast down to a given reference type
-  - `br_on_cast $l $t : [(ref null ht) (rtt n? $t)] -> [(ref null ht)]`
-    - iff `ht <: data` or `ht <: func`
-    - and `(type $t) <: ht`
-    - and `$l : [(ref $t)]`
+* `br_on_cast <labelidx>` branches if a value can be cast down to a given reference type
+  - `br_on_cast $l : [t (rtt n? $t')] -> [t]`
+    - iff `$l : [t']`
+    - and `t <: dataref` or `t <: funcref`
+    - and `(ref $t) <: t'`
```
manoskouk, line 406: I think there are some aliased variables here. It should be:
`br_on_cast $l : [t1 (rtt n? $t2)] -> [t1]`
- iff `$l : [t3]`
- and `t1 <: (ref null data)` or `t1 <: funcref` (instead, `t1 <: anyref` if we go with the above suggestion)
- and `(ref $t2) <: t3`

proposals/gc/MVP.md
```diff
@@ -379,29 +387,27 @@ TODO: Add the ability to generate new (non-canonical) RTT values to implement ca
 
 RTT-based casts can only be performed with respect to concrete types, and require a data or function reference as input, which are known to carry an RTT.
 
-* `ref.test <typeidx>` tests whether a reference value's [runtime type](#values) is a [runtime subtype](#runtime) of a given RTT
-  - `ref.test $t : [(ref null ht) (rtt n? $t)] -> [i32]`
-    - iff `ht <: data` or `ht <: func`
-    - and `(type $t) <: ht`
+* `ref.test` tests whether a reference value's [runtime type](#values) is a [runtime subtype](#runtime) of a given RTT
+  - `ref.test : [t' (rtt n? $t)] -> [i32]`
+    - iff `t' <: dataref` or `t' <: funcref`
```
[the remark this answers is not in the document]
rossberg: Ah, no, the purpose of introducing the separate classification instructions actually was to simplify the test/cast instructions, so that they only operate on references that are known to have an RTT. Otherwise there function would overlap with the classification instructions. See #150 for more context.

You are right about the null thing, fixed.

rossberg (commented):
Thanks, I pushed a quick fix to the typos you point out.

proposals/gc/MVP.md
rossberg: There was a typo in the last line, which should have been `(ref $t')` (fixed). But I think the others are correct, considering that `$t'` and `t'` are different names and in different categories.

PR #192, opened 2021-02-04
Allow extra args on branches

rossberg:
Fixes inconsistency with `br_on_null`, as [pointed out](https://github.com/WebAssembly/function-references/issues/45#issuecomment-772678955) by @kripken.

### DIFF
--- a/proposals/gc/MVP.md
+++ b/proposals/gc/MVP.md
@@ -325,25 +325,25 @@ Tentatively, support a type of guaranteed unboxed scalars.
   - `ref.is_i31 : [anyref] -> [i32]`
 
 * `br_on_func <labelidx>` branches if a reference is a function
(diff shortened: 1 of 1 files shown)


kripken (commented):
Seems reasonable, thanks.

I sort of prefer the other way, actually - it would keep things even simpler to not allow extra values. Do you have a sense of whether they would be useful? (FWIW, in code I've seen I doubt they would, but that's mostly LLVM.)

rossberg:
I'd be fine the other way as well, though one could argue that's a rather artificial restriction in a stack machine.

FWIW, @manoskouk suggested this generalisation earlier (#183), and I was skeptical myself. But being consistent seems desirable.

kripken (approved):
Fair points. I agree this makes things more consistent which is good. Thanks!

PR #194, opened 2021-02-08
Fix metavariables in rtt.sub spec

tlively:


proposals/gc/MVP.md
```diff
@@ -376,7 +376,7 @@ Note: The `br_on_*` instructions allow an operand of unrelated reference type, e
 
 * `rtt.sub <typeidx>` returns an RTT for `typeidx` as a sub-RTT of a the parent RTT operand
   - `rtt.sub $t : [(rtt n $t')] -> [(rtt (n+1) $t)]`
-    - iff `t2 <: t1`
+    - iff `t <: t'`
```
tlively, line 379: Should there be an additional requirement that `t != t'`, or is intended that the same type can be used in successive rtts in a chain?

kripken: A possible use of `t == t'` is that `(ref.canon $foo) != (ref.sub $foo (ref.canon $foo))`. But I do not know if that is useful in practice.

RossTate: If rtt depth is supposed to be able to correspond to class-inheritance depth, then `t == t'` would happen when a subclass does not add new fields.

rossberg: Ah, thanks, but the symbolic names are merely indices now. So I think it needs to say this:
```suggestion
    - iff `(type $t) <: (type $t')`
```

PR #213, opened 2021-05-10
Add negated versions of br_on_* instructions

rossberg:
As a result of the discussion on WebAssembly/function-references#45, this adds inverse versions of the `br_on_*` and `br_cast` instructions as well.

### DIFF
--- a/proposals/gc/MVP.md
+++ b/proposals/gc/MVP.md
@@ -322,12 +322,25 @@ Tentatively, support a type of guaranteed unboxed scalars.
 * `ref.is_i31` checks whether a reference is an i31
   - `ref.is_i31 : [anyref] -> [i32]`
 
+* `br_on_non_null <labelidx>` branches if a reference is not null
+  - `br_on_non_null $l : [t0* (ref null ht)] -> [t0*]`
+    - iff `$l : [t0* (ref ht')]`
+    - and `ht <: ht'`
+  - passes operand along with branch, plus possible extra args
+
 * `br_on_func <labelidx>` branches if a reference is a function
   - `br_on_func $l : [t0* t] -> [t0* t]`
     - iff `$l : [t0* t']`
     - and `t <: anyref`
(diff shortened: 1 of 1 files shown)


proposals/gc/MVP.md
```diff
@@ -322,27 +322,54 @@ Tentatively, support a type of guaranteed unboxed scalars.
 * `ref.is_i31` checks whether a reference is an i31
   - `ref.is_i31 : [anyref] -> [i32]`
 
+* `br_on_non_null <labelidx>` branches if a reference is not null
+  - `br_on_non_null $l : [t0* (ref null ht)] -> [t0*]`
```
kripken, line 326: for symmetry (with `br_on_null`, and the others here) should this return the input perhaps?
```suggestion
  - `br_on_non_null $l : [t0* (ref null ht)] -> [t0* (ref null ht)]`
```

rossberg: Well, br_on_null does not pass on the null reference to the label, so the symmetry is that br_on_non_null inversely does not return it. We could have designed br_on_null differently, but it's too late for that.

Is there a use case where you'd actually use the null?

kripken: I have no specific use case in mind. Yeah, I suppose it's not useful. And good point about `br_on_null` not passing the value.

proposals/gc/MVP.md
```diff
@@ -501,6 +537,7 @@ The opcode for heap types is encoded as an `s33`.
 | 0xfb40 | `ref.test $t` | `$t : typeidx` |
 | 0xfb41 | `ref.cast $t` | `$t : typeidx` |
 | 0xfb42 | `br_on_cast $l` | `$l : labelidx` |
+| 0xfb42 | `br_on_cast_fail $l` | `$l : labelidx` |
```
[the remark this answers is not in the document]
rossberg: Thanks, fixed.

PR #242, opened 2021-09-14
Sync instruction mnemonics

rossberg:
This updates the MVP overview to match the instruction names actually used in the test suite and interpreter.

### DIFF
--- a/proposals/gc/MVP.md
+++ b/proposals/gc/MVP.md
@@ -246,12 +246,12 @@ This can compile to machine code that (1) reads the RTT from `$x`, (2) checks th
 
 #### Structures
 
-* `struct.new_with_rtt <typeidx>` allocates a structure with RTT information determining its [runtime type](#values) and initialises its fields with given values
-  - `struct.new_with_rtt $t : [t'* (rtt n $t)] -> [(ref $t)]`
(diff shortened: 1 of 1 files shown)


jakobkummerow:
It's true that I originally suggested these renamings for simplification, back when it seemed like those would be the only `.new` instructions.

But nowadays we have [subgroup consensus](https://github.com/WebAssembly/meetings/blob/main/gc/2021/GC-08-24.md#discussion-evaluating-nominal-types-ben-titzer-20-minutes) to officially evaluate nominal types, which requires a second set of RTT-less `.new` instructions. For disassembly/debugging purposes, both opcodes (e.g. `0xfb01` and `0xfb07` could be translated to the same human-readable `struct.get`, but since the .wat format is also used as an assembler/interchange format by current toolchains (e.g. produced by J2CL, consumed by Binaryen), it is useful to have an unambiguous distinction between the two instructions in the text format.

So at this time, to accomplish our goals, we should maintain the distinction between:
- `struct.new`, `0xfb07`, without RTT on the stack, with static type
- `struct.new_with_rtt`, `0xfb01`, with RTT on the stack, possibly without static type (#241)
(And analogously for `new_default` and `array.new*`.)

In the future, we will decide whether both instructions will be kept or only one of them; in the latter case, whichever is kept can be renamed to just `struct.new` for simplicity.

tlively:
Closing this because it looks like it is no longer applicable.

Issue #217, opened 2021-05-18
Isorecursion seems too weak


RossTate:
Here are two class hierarchies that isorecursion (as just presented) seems to not support (but which nominal types support):
```
class A { A foo(B); }
class B extends A { override B foo(B); }
```

```
class X { X bar(); }
class Y extends X { Z z; override Y bar(); }
class Z { Y y; }
```

tlively:
Once the slides from this morning's meeting are posted, it would be great if you could elaborate on how the system as presented fails in these cases. I tried to reason through it myself, but my memory/understanding of the presented subtyping details is too spotty :)

rossberg:
As far as I can see, these examples are expressible just fine, provided we support nested recursive types as I alluded to (I didn't present the full Wasm syntax/rules for that, but you can extrapolate from the standard rules for iso-recursive types). All you have to do is wrap classes into another layer of type recursion:
```
;; $AB.$B <: $AB.$A
type $AB = (rec
  $A = (rec
    $inst = struct (ref $AB.$A.$vt)
    $vt = struct (ref $AB.$A.$foo)
    $foo = func (ref $AB.$A.$inst) (ref $AB.$B.$inst) -> (ref $AB.$A.$inst)
  )
  $B = (rec
    $inst = struct (ref $AB.$B.$vt)
    $vt = struct (ref $AB.$B.$foo)
    $foo = func (ref $AB.$A.$inst) (ref $AB.$B.$inst) -> (ref $AB.$B.$inst)
  )
)
```
```
;; $XYZ.$Y <: $XYZ.$X
type $XYZ = (rec
  type $X = (rec
    $inst = struct (ref $XYZ.$X.$vt)
    $vt = struct (ref $XYZ.$X.$bar)
    $bar = func (ref $XYZ.$X.$inst) -> (ref $XYZ.$X.$inst)
  )
  $Y = (rec
    $inst = struct (ref $XYZ.$Y.$vt) (ref $XYZ.$Z.$inst)
    $vt = struct (ref $XYZ.$Y.$bar)
    $bar = func (ref $XYZ.$X.$inst) -> (ref $XYZ.$Y.$inst)
  )
  $Z = (rec
    $inst = struct (ref $XYZ.$Z.$vt) (ref $XYZ.$Y.$inst)
    $vt = struct
  )
)
```
[...]

PR #243, opened 2021-09-15
Define iso-recursive type semantics

rossberg:
This is a detailed definition of the iso-recursive type semantics with explicit subtyping that I proposed in the recent meeting. I tried to keep the type representation as "natural" as possible, e.g. avoiding explicit type variables or substitutions except as auxiliary notions in the definition of type equivalence . Please let me know if something is unclear (or wrong).

I intend to implement this in the interpreter next.


proposals/gc/MVP.md
```diff
@@ -97,20 +118,167 @@ New abbreviations are introduced for reference types in binary and text format,
...
+* a plain type definition is valid if its `deftype` is at its type index
+  - `<deftype> ok` and extends the context with `<deftype>`
+    - iff `<deftype> ok($t)` where `$t` is the next unused (i.e., current) type index
+
+* a recursive type definition is valid if its types are valid under the context containing all of them
+  - `rec <deftype>* ok` and extends the context with `<ctxtype>*`
+    - iff `<deftype>* ok($t)` under the extended context(!)
+    - where `x` is the next unused (i.e., current) type index
+    - and `N = |<deftype>*|-1`
+    - and `<deftype'>* = <deftype>*[$t:=rec $t, ..., $t+N:=rec $t+N]`
```
fgmccabe, line 170: Should this read:
    - and `<deftype'>* = <deftype>*[$x:=rec $x, ..., $x+N:=rec $x+N]`
?

proposals/gc/MVP.md
```diff
@@ -97,20 +118,167 @@ New abbreviations are introduced for reference types in binary and text format,
...
+    - and `<ctxtype>*  = (rec <deftype'>*).0, ..., (rec <deftype'>*).N`
+  - Note: `<deftype'>*` marks all recursive occurrences of type indices from within this group with `rec`; this is expressed here by a validation-time substitution, but an implementation could insert the annotations on the fly during decoding.
+  - Because rec type indices `rec $t`cannot occur in the source, the following are invariants that are established by these rules:
+    - (1) rec indices only occur in the context and inside a `(rec <deftype>*)`,
+    - (2) they only refer to indices from the same recursive group,
+    - (3) all internal indices within a recursive group are marked with rec,
```
fgmccabe, line 176: Does this imply that with rec <idx>, <idx> is in the range 0..N where there are N definitions in the rec <deftype> group?

rossberg:
I figured that the formulation was a bit hacky and obscure after all, so I replaced it was something more "standard" and simultaneously more confined to the definition of type equivalence. It should be semantically equivalent to the earlier formulation, but cleaner.

tlively:
Thanks for posting this, @rossberg, and sorry for the delay in responding to it. I still don't completely understand all the details here, so I could use help in answering a few questions.

1. We previously had a [concrete example](https://github.com/WebAssembly/gc/issues/217#issuecomment-847239533) where isorecursion was unable to express some Java subtyping. Is that example expressible in this new system? Can we be confident that there are no other expressiveness problems?

2. We have found in our optimization work that it is important in Binaryen to be able to have different types for different nominal source types that are otherwise structurally equivalent. For example, it is important that we have different types for two different but structurally equivalent subclasses of `Object`. Does this system allow us to define two distinct but structurally equivalent types? Does putting the two structurally equivalent definitions in the same rec group define two distinct types, perhaps?

3. Is the expectation that whole-program compilers will use just a single large rec group in their output?

4. The goal of using structural typing is to let source-level structural types to correspond to Wasm types that can be defined separately in various modules without needing any imports to coordinate type identity. How would that look concretely in this system? Would schemes for separate compilation put each connected component of source-level structural types into its own rec group that is duplicated along with its dependencies in every module that uses one of those types?

5. What does the canonicalization algorithm for this type system look like?

proposals/gc/MVP.md
```diff
@@ -565,3 +750,215 @@ See [GC JS API document](MVP-JS.md) .
...
+--------------------------
+C |- sub x* st ok(x')
+
+C |- dt ok(x)
+C |- dt'* ok(x+1)
+-------------------
+C |- dt dt'* ok(x)
+```
+
+#### Type Definitions (`C |- <deftype>* -| C'`)
```
tlively, line 814: I'm not familiar with this reverse turnstile syntax. What does it mean?

proposals/gc/MVP.md
```diff
@@ -565,3 +750,215 @@ See [GC JS API document](MVP-JS.md) .
...
+C |- st ok
+(C |- st <: expand(C(x)))*
+(x < x')*
+--------------------------
+C |- sub x* st ok(x')
+
+C |- dt ok(x)
+C |- dt'* ok(x+1)
+-------------------
+C |- dt dt'* ok(x)
```
tlively, line 811: What is this rule doing? I don't understand what `dt` and `dt'*` are.

proposals/gc/MVP.md
```diff
@@ -565,3 +750,215 @@ See [GC JS API document](MVP-JS.md) .
...
+------------------------------------
+C |- func t1* t2* == func t1'* t2'*
+
+(C |- ft == ft')*
+----------------------------
+C |- struct ft* == struct ft'*
+
+C |- ft == ft'
+--------------------------
+C |- arrat ft == array ft'
```
tlively, line 899:
```suggestion
C |- array ft == array ft'
```

proposals/gc/MVP.md
```diff
@@ -565,3 +750,215 @@ See [GC JS API document](MVP-JS.md) .
...
+
+#### Field Types (`C |- <fldtype> == <fldtype'>`)
+
+```
+C |- t == t'
+--------------------
+C |- mut t == mut t'
+```
+
+#### Structural Types (`C |- <strtype> == <strtype'>`)
```
tlively, line 885: What is the difference between a structural type and a defined type?

proposals/gc/MVP.md
```diff
@@ -565,3 +750,215 @@ See [GC JS API document](MVP-JS.md) .
...
+
+#### Type Indices (`C |- <typeidx> <: <typeidx'>`)
+
+```
+C |- x == x'
+------------
+C |- x <: x'
+
+unroll(C(x)) = sub (x1* x'' x2*) st
+C |- x'' <: x'
```
tlively, line 921: Would there be any downside to putting subtype depth into the syntax (optionally?) so subtype checks can be constant time rather than linear like we currently do with RTTs?

proposals/gc/MVP.md
```diff
@@ -97,20 +106,187 @@ New abbreviations are introduced for reference types in binary and text format,
...
+
+* `rec.<i>` is a new form of type index
+  - `typeidx ::= ... | rec.<i>`
+
+This form is only used during equivalence checking, to identify and represent "back edges" inside a recursive type. It is merely a technical device for formulating the rules and cannot appear in source code. It is introduced by the following auxiliary meta-function:
+
+* Rolling a context type produces an _iso-recursive_ representation of its underlying recursion group
+  - `tie($t)                    = tie_$t(<ctxtype>)`  iff `$t = <ctxtype>`
+  - `tie_$t(<subtype>)          = <subtype>`
+  - `tie_$t((rec <subtype>*).i) = (rec <subtype>*).i[$t:=rec.0, ..., $t'+N:=rec.N]` iff `$t' = $t-i` and `N = |<subtype>*|`
```
tlively, line 207:
```suggestion
  - `tie_$t((rec <subtype>*).i) = (rec <subtype>*).i[$t':=rec.0, ..., $t'+N:=rec.N]` iff `$t' = $t-i` and `N = |<subtype>*| - 1`
```
IIUC

RossTate:
Recently we have had a number of discussions establishing consensus on what we want for specifically the MVP. In light of the outcomes of those discussions, could you articulate why this variant on isorecursive types should be an MVP feature rather than a Post-MVP feature.

conrad-watt:
@RossTate given that we agreed to prototype both a nominal and a structural system as MVP candidates, I'm not sure what the motivation would be for requiring such a justification right now, unless you're suggesting that an alternative structural proposal would be a more viable MVP candidate?

RossTate:
According to the meeting notes, the following is what we agreed to:

> Poll, we will evaluate nominal types by adding the minimum delta necessary (types and instructions) to the current MVP and tools.

The changes in this PR are substantial and will require significant work to implement. As with anything requiring substantial resources, a motivation should be provided.

conrad-watt:
Then the poll was poorly phrased given the spirit of what was decided, and if you feel it's necessary we can secure a revised wording in a future meeting. Since we have a group understanding that equi-recursive canonicalisation is not viable, making a decision based on an evaluation of the current MVP against a nominal proposal is essentially the same as selecting the nominal proposal. Given the conversations that took place in the meeting, I don't think many people interpreted the poll in this way.

titzer:
It was my understanding that we would pursue both structural and nominal types, and the only viable option (due to the performance issues with equi-recursive types) was an iso-recursive formulation. So I have been working under the assumption that we'd be doing this change. So I generally support this PR, though I have no dug into the details.

tlively:
Speaking for Google folks, we've been implementing the system described in the [milestone 4 prototype spec](https://docs.google.com/document/d/1afthjsL_B9UaMqCA5ekgVmOm75BVFu6duHNsN9-gnXw/edit#heading=h.y0pmx2oa4v0b) to fulfill the intent of that poll. We're also interested in the system in this PR as a potential next step because the milestone 4 spec still uses equirecursive types for its structural component and we will want a better structural component in the final MVP system if there will be a structural component at all. The first step to investigate this system further is discussing it to see if any issues come up that would make implementing it impractical or unnecessary, so I am happy to have this PR open to enable that discussion.

proposals/gc/MVP.md
tlively: I see, a `deftype` is a type section entry defined as `<subtype> | rec <subtype>*`, so type section entries can either be on their own (like the current function types) or inside a `rec` group. A `subtype` is a single type definition, whether it is on its own or in a `rec` group.

rossberg:
[Sorry for not replying earlier, was out sick.]

> > 1. We previously had a [concrete example](https://github.com/WebAssembly/gc/issues/217#issuecomment-847239533) where isorecursion was unable to express some Java subtyping. Is that example expressible in this new system? Can we be confident that there are no other expressiveness problems?
> 
> I think that example works in this new system because subtyping is precisely according to the declared supertypes [...] This is exactly how subtyping works in our nominal implementation.

Right.

> One potential problem I can think of is when the ordering constraint on declared supertypes comes into conflict with the adjacency constraint of types within a `rec` group. For example if you had `A <: B <: C` and `A` and `C` were mutually recursive, but for some reason `B` had to be in a different `rec` group. I can't think of any truly concrete examples of that problem, but I could imagine an artificial situation where `A` and `C` should be shared structurally between modules but `B` should not be shared structurally.
> If we do find (or believe there exist) concrete instances of this problem, the simplest fix would be to disallow cyclic supertype chains without requiring an ordering constraint. That's the approach we've implemented in Binaryen's nominal implementation and the approach described in the milestone 4 prototype.

Right, there are alternative ways to impose the non-cyclicity constraint, but this is the simplest, and I can't think of any scenario where it would be a serious problem.

Note that in your example, B could not be pulled out of the recursion group no matter what. The definition of C depends on B depends on A, so they are all mutually recursive.


> > 1. We have found in our optimization work that it is important in Binaryen to be able to have different types for different nominal source types that are otherwise structurally equivalent. For example, it is important that we have different types for two different but structurally equivalent subclasses of `Object`. Does this system allow us to define two distinct but structurally equivalent types? Does putting the two structurally equivalent definitions in the same rec group define two distinct types, perhaps?
> 
> Yes, I believe that two structurally equivalent definitions in the same rec group define distinct types. The reason is that recursive types are equivalent only when their `rec` groups are pointwise equivalent _and the projected indices are the same_. A whole-program compiler can use this property to exactly recover nominal behavior by putting all of its types in a single `rec` group.

Yes, the degenerate case of just using one whole-program recursive group essentially emulates fully nominal behaviour.

That said, I'm confused why an optimisation would depend on distinguishing equivalent representations. Can you elaborate? This is the first time I'm hearing of this problem, I believe.


> > 1. Is the expectation that whole-program compilers will use just a single large rec group in their output?
> 
> If they do, they will get the current nominal behavior but won't be able to structurally share types. This is fine because these compilers see the whole program, so they should feel free to do this.

I don't necessarily expect them to do that, but of course they could, if it avoids some hassle for them.


> > 1. The goal of using structural typing is to let source-level structural types to correspond to Wasm types that can be defined separately in various modules without needing any imports to coordinate type identity. How would that look concretely in this system? Would schemes for separate compilation put each connected component of source-level structural types into its own rec group that is duplicated along with its dependencies in every module that uses one of those types?

Yes. This duplication is generally unavoidable anyway, because at the import boundaries each module has to repeat the definition of any imported type it uses, even under a hypothetical fully nominal semantics. Otherwise, validation, compilation, and link-time checking wouldn't work under Wasm's module system. However, a nominal or equi-recursive semantics might get away with keeping some types in a recursion group fully abstract in some import cases, while iso-recursive does not. Not clear how relevant this case would be in practice, though. But it certainly is an incentive to keep recursion groups small when doing separate compilation.

> A compiler that does want to share source-level structural types structurally would have to place the shared structural types in their own `rec` group(s). Compilers intending to share types structurally would need to agree on a convention for grouping and ordering the shared types (and all of their component types) for the sharing to work out.

Yes, producers have to canonicalise type definitions suitably. That is already the case in some places, e.g., with the order of fields in a struct, or the order of cases in a source-level variant type (when lowering it to int tags).

FWIW, the main use case here is linking multiple modules produced by the same compiler, so the type canonicalisation ought to be implicitly consistent. Multiple different compilers for the same language interacting is not a use case we have seriously discussed thus far. If they would not want to go through interface types, they would of course have to agree on and codify their low-level data layout strategy. But that is a fairly hypothetical scenario, I think – I'm not aware of such degree of compatibility between high-level language compilers for native code elsewhere.

> > 1. What does the canonicalization algorithm for this type system look like?
> 
> Each `rec` group is represented as a single finite syntax tree and the set of all `rec` groups is organized into a dependence tree. The dependence tree is acyclic because mutually recursive types must always be in the same `rec` group (otherwise they wouldn't be valid). `rec` groups are deduplicated bottom-up in the dependency graph by hash-consing their structures and comparing their dependencies (which must have already been canonicalized).

Exactly. That is the main property of the iso-recursive semantics.

> ### A problem
> Consider module 1 defining structural tuple types `A` and `C` such that `C <: A`. Module 1 must "nominally" declare that `A` is `C`'s supertype. Then consider module 2 using `A` and `C` and also defining another structural tuple type `B` such that we should have `C <: B <: A`. But `C` was declared to have `A` as a supertype, so any code in module 2 depending on `C <: B` will not validate. Or alternatively, module 2 could declare `C <: B` as intended, but then `C` in module 2 would be a distinct type from `C` in module 1 because type equivalence requires declared supertypes to match.
> So it seems that this system cannot directly express shared structural types when those structural types have subtyping, which I think is the case for at least function types in most relevant languages. I'm struggling to think of workarounds that would not also work in a simpler, purely nominal system.

Yes, with pre-declared subtyping, you have the same limitations in this regard as any such system, including a nominal one.

> Given this limitation, are the benefits of this system over a purely nominal system worth the extra complexity?

The benefit of structural typing is independent of that. When compiling nominal source languages, this obviously is a non-issue. When compiling structural source languages that do not have source-level subtyping, you also have no issue. Such languages might still make internal use of subtyping hierarchies, but these tend to be very shallow and well-layered (such as lowering closure types). That is the main use case this is trying to address.

The only scenario where the limitation would become relevant would be lowering a source language with structural subtyping to Wasm in a direct, non-coercive manner. Such non-coercive compilation obviously won't fly with pre-declared subtyping, no surprise there. That is the price we are willing to pay, since such languages are rare, and non-coercive compilation for them would likely run into other blockers anyway (e.g., wrt to generics).

If this ever becomes a practical concern, the explicit, zero-cost coercion operator I mentioned in my presentation could be a way out. This possibility was pointed out to me by Sam Lindley.

proposals/gc/MVP.md
rossberg: Unless you consider subtype constraints on type imports (which is not included here), the depth is implicit in the definition, since all supertype definitions are fully transparent.

For imports, it may be worth considering some restrictions, e.g., putting the depth in the constraint, but I'd look at that separately.

proposals/gc/MVP.md
rossberg: Oh, actually, that should be `st` and `st*`. Will fix.

proposals/gc/MVP.md
tlively: Makes sense, thanks.

tlively:
> That said, I'm confused why an optimisation would depend on distinguishing equivalent representations. Can you elaborate? This is the first time I'm hearing of this problem, I believe.

@kripken has implemented "constant field propagation" in Binaryen that does a whole-program type-based analysis to discover what values a struct field might have, and if there is just one, propagates that value directly to uses of that field. Combined with other optimizations this implements devirtualization and gives a 40% speedup on J2CL output. However, to be effective it requires that nominally distinct but structurally identical types are in fact distinct, otherwise their different assigned field values would defeat the optimization.

> The benefit of structural typing is independent of that. When compiling nominal source languages, this obviously is a non-issue. When compiling structural source languages that do not have source-level subtyping, you also have no issue. Such languages might still make internal use of subtyping hierarchies, but these tend to be very shallow and well-layered (such as lowering closure types). That is the main use case this is trying to address.
> 
> The only scenario where the limitation would become relevant would be lowering a source language with structural subtyping to Wasm in a direct, non-coercive manner.

Can you give some concrete examples of languages that have structural source types and do or do not have structural subtyping? As far as I can tell, OCaml for example has structural subtyping of function types, and so would Java if lambdas weren't compiled down to nominal classes. I expect most languages would be similar and therefore would run into this issue.

> If this ever becomes a practical concern, the explicit, zero-cost coercion operator I mentioned in my presentation could be a way out. This possibility was pointed out to me by Sam Lindley.

I would be interested in hearing more about that as well!

kripken:
> @kripken has implemented "constant field propagation" in Binaryen that does a whole-program type-based analysis to discover what values a struct field might have, and if there is just one, propagates that value directly to uses of that field. Combined with other optimizations this implements devirtualization and gives a 40% speedup on J2CL output. However, to be effective it requires that nominally distinct but structurally identical types are in fact distinct, otherwise their different assigned field values would defeat the optimization.

Checking the history, it looks like that presentation was [on 8/24](https://github.com/WebAssembly/meetings/blob/main/gc/2021/GC-08-24.md#presentation-update-on-binaryen-optimizations-alon-zakai-20-minutes) ([slides](https://github.com/WebAssembly/meetings/blob/main/gc/2021/presentations/2021-08-24-zakai-optimization-update.pdf)).

RossTate:
>> If this ever becomes a practical concern, the explicit, zero-cost coercion operator I mentioned in my presentation could be a way out. This possibility was pointed out to me by Sam Lindley.
>
> I would be interested in hearing more about that as well!

This is where you have an instruction (or instruction sequence) that are all erased but show the type-checker why values of one type necessarily belong to another type. This is the approach that most typed assembly languages used. However, the only team to actually manage to do this for a real language found that it was so cumbersome and required so much expertise to generate that they believed it would preclude typed assembly languages from ever being adopted. (They then recruited me to solve the problem.)

askeksa-google:
> So right now, I am not aware of a relevant language that would (a) have source level structural subtypes and (b) lower them directly to equally expressive representation-level subtypes.

FWIW, Dart has structural subtyping for its function types. The subtyping relation is based both on the types of the parameters/return and on the structure of optional parameters (i.e. a signature with an optional parameter is a subtype of one which does not have that optional parameter and is otherwise identical).

As with other Dart types, upcasts are implicit, and both downcasts (`as` expressions) and type tests (`is` expressions) do a full subtype check at runtime. Calls of values with type `dynamic` or `Function` (function with unspecified signature) dynamically check adherence to the signature based on the runtime types of the arguments.

Would it perhaps make sense in this proposal to have function type subtyping be implicit and structural while structs and arrays are explicit? Since we don't have a cast instruction for function types, their subtyping relation is only ever used statically anyway.

RossTate:
I have found issues with this proposal and separate compilation of object-oriented languages. These issues are in fact already present in the Wob compiler (and are not addressed by the changes in this PR).

@tlively @conrad-watt @titzer Please advise: are we or are we not evaluating this PR on separate compilation? If we are, then I'll go into these issues, and I will also present a design that addresses the concerns raised with separate compilation of OCaml/SML in a more general manner while keeping MVP subtyping constant time (and without using string mangling) and supporting much more extensibility. If we are not, then please provide some sense of how we plan to evaluate this PR with respect to what has been implemented.

---

@askeksa-google In weak mode, Dart's function subtyping is unsound and so its function subtyping cannot be supported by wasm's function subtyping. In strong mode, all function types will have to be uniform in order to support generics. (There are also two larger issues. First, adding such function subtyping would make subtyping undecidable in order to provide even basic support for generics. Second, surprisingly function types seem to have no performance benefits once dynamically typed function calls are implemented well, whereas the flexibility of dynamically typed function calls seem to enable them to provide better performance for things like interface methods or `dynamic` calls.)

kripken:
@rossberg 

> So, I assume this has do some type analysis pass first? If so, wouldn't it still be possible to optimistically assume that multiple structurally identical types are separate, and merge them only when type analysis determines that they are actually mixed?

Optimizations in Binaryen work on the wasm after loading it, and canonicalization happens during load. So the optimization would never get a chance to see that merged types began as separate ones before canonicalization.

tlively:
@RossTate, yes, we are evaluating this PR on its suitability for separate compilation. Although no one present in the last subgroup meeting thought that separate compilation was an important issue for the MVP, @rossberg has made it clear that he does and that this PR is trying to address the problems of separate compilation of structural source types. If this system does not achieve that goal, it would be good to know that as early as possible.

RossTate:
Thanks for the clarification. Here are three issues that come to mind.

---

According to its documentation, and from what I can tell of its implementation strategy, the Wob compiler does not support dynamic casts. The implementation only provides infrastructure for implementing casts with rtts (the old implementation uses `rtt.sub`, but it seems the new implementation will use `rtt.canon` and rely on the fact that there's now an explicit subtyping hierarchy). The issue with this is that if two distinct classes have the same structure then they'll have the same rtt, which means that using that rtt will make casts succeed that should have failed.

Using this PR, you can solve that problem in a whole-program setting. You simply place all classes in one big recursion group. That will cause distinct classes to be at distinct indices and therefore have distinct rtts. But with separate compilation, you cannot add new classes into an existing recursion group, and so instead you have to manually implement casting.

Yes, you can do this manual casting yourself, but you can also do canonicalization yourself. But the latter only affects loading, whereas the former affects representation/size/execution.

---

Let's suppose Wob didn't have generics (e.g. like the JVM or early Java), but it still has tuples (which already are not always generic anyways, e.g. a `(Float, Float)` cannot be given to `func fst<X,Y>(p : (X,Y)) : X`). In that case, you could represent a tuple `(Foo, Bar)` as a `ref (struct (ref $Foo) (ref $Bar))`. But `(Foo, Bar)` is also a subtype of `(Foo, Object)` and `(Object, Bar)`, both of which are subtypes of `(Object, Object)`. So this means we'd have to enable the extension to multiple explicit supertypes to handle proper covariant tuples. It also means that if someone writes `<Bar, Bar, Bar, Bar, Bar>`, and `Bar` has n supertypes (either directly or indirectly), then the generator has to enumerate all n<sup>5</sup> supertypes of the tuple.

Matters get worse when you consider structural types like (covariant) `List<Foo>`. Because the structural definition of `List` is recursive, this PR offers no way for someone to make `List<Foo>` a subtype of `List<Object>`.

I know how to support these examples in a Post-MVP by using nominal types, but I do not know how to support them (decidably) with structural types. (Similarly, I know how to support OCaml generics by using nominal types in a Post-MVP, but I do not know how to do so with structural types.)

---

Now, ignore both subtyping and generics for a bit, but suppose our structural language has exceptions. When you throw an exception, you specify an arbitrary number of values for the payload. When you catch an exception, you specify an arbitrary number of variables for the payload, each with a type. In order to actually catch a thrown exception, the catch's arity and types must match that of the throw. (There's no subtyping, so the match can be exact.)

With whole-program compilation, you can declare a different exception tag for each arity and type list used in the program. Each `throw` and `catch` then uses the designated tag for the given arity and type list, and this recreates the semantics of the program.

However, with separate compilation you have the issue that there's an infinite number of possible exception arities and type lists, and you have to make sure that separately compiled modules use the same tag for the same arity and type list. This same issue actually arises for rtts, which is why this design is completely useless if `rtt.canon` does not exist. Should we add a way to get canonical tags for exceptions? But even if we did, it wouldn't properly solve the problem because it would canonicalize according to *wasm* types rather than *surface* types. (Note that the same issue arises even for OCaml if one were to implement functions using call tags rather than a casting hierarchy, which we're accruing evidence would perform better.)

Yes, you can solve this issue by boxing exceptions and using dynamic tests, or you can solve it by finitely approximating exceptions, but the same is true for the examples motivating this PR.

---

All this is to illustrate that this PR does not add support for separate compilation of structural languages. Rather, it bakes in support for a specific compilation and loading strategy for specifically OCaml/SML, and at the cost of many other languages (possibly including OCaml/SML if ever we add generics). Support for this compilation and loading strategy could also be provided by a general-purpose (and declarative) canonicalization feature that could handle types/rtts/tags/..., and if one were to allow modules to use keys besides just wasm types (e.g. using constants representing surface-level types according to some convention developed by the compiler), then this canonicalization could be much more aligned with what languages actually need (and likely more efficient).

I am happy to provide more examples of issues or go into more depth on alternative solutions, but I am worried about how hypothetical everything is. Unless the group wants to really bring separate compilation into the MVP, I do not believe we will be able to properly evaluate these options until Post-MVP. Furthermore, I do not see a way in which the current design precludes adding the features of this PR in a Post-MVP.

tlively:
Thanks, @RossTate. To summarize and respond to each issue:

1. Dynamic casts have to be implemented "manually" because distinct source types may have identical underlying structural types and the trick of placing them in the same `rec` group does not work across modules.

    - This is true of all the structural proposals so far. Anecdotally, the J2CL has been very happy to have this taken care of by nominal typing, but we don't have data on the actual size or performance impacts of manually implementing casting.
  
2. Tuple types with subtyping form subtype semi-lattices rather than trees, requiring either the multiple supertypes extension or a separate subtype for each supertype with a coercion operator inserted where necessary.

    - This does seem complicated, but it's also worth noting that the currently implemented nominal system has this same problem and we have not discussed it at all or proposed any solutions. It would be good see a prototype of these potential solutions working, perhaps in Wob. By the way, what real-world languages have this kind of tuple subtyping?

3. Exception tags are nominal, so they have to be coordinated in a nominal fashion, even when their payloads are structural types.

    - This has also been true of all the structural proposals so far. It's also true of the current nominal prototype, except that the nominal prototype would additionally coordinate everything else nominally as well. In practice the only way exception tags have been used so far is with a uniform representation for payloads, so this has not been a problem. Like for "manual" dynamic casting, we don't have data on the performance penalty we are paying by manually testing exception payload types.

Overall the severity of these problems can only be assessed by prototyping separate compilation and the reason they aren't problems for the nominal system is that the nominal system explicitly declines to try to solve them, which is fine.

To evaluate the utility of this hybrid system relative to the nominal system we need to determine what it supports that the nominal system does not support. In other words, what is the set of (real-world) languages that this proposal would allow to be separately compiled without the kind of external coordination that would be necessary in the nominal system? To that end, @rossberg, would you be able to list a few languages that you believe this proposal would enable to be separately compiled? If we can agree on what languages this proposal does support and what languages this proposal does not support, we will be in a good place.

RossTate:
Thanks for the thorough summary, @tlively!

RossTate:
> By the way, what real-world languages have this kind of tuple subtyping?

Sorry, I realized I forgot to answer this question. Scala and Ceylon come to mind (though sadly the latter is no longer maintained). C# doesn't support covariant tuples because it wants a `Tuple<double,byte>` to be representable without boxing the `double` or `byte` despite the fact that both are subtypes of `object`. (It does support variant delegates, though, because you can support specialized representation in delegates by using techniques like call tags.) Note also that generics come up a lot in this issue of structural types, so I expect it will be hard to properly evaluate support for structural types in a system without support for generics.

This discussion of tuples reminds me that every major language, for one reason or another, seems to support tuples either up to a certain fixed size or simply using arrays. Scala tuples go up to 22 because it has to generate a v-table for each tuple. Haskell98 tuples go up to 15 (though GHC recently goes up to 62) because it needs to generate type classes. C# goes up to 8, again because it has to generate v-tables. OCaml represents tuples as arrays in order to support efficient polymorphic structural equality/comparison. But maybe SML doesn't need to do this? Anyways, the claim that WebAssembly needs to support linking of arbitrary-sized tuples is contrary to common practice. Major languages ship support for fixed-size tuples with either the runtime or the standard library and expects code using tuples to explicitly link with the runtime or standard library for the sizes used in order to get access to their common functionality (or simply represent tuples as arrays). Similar issues come up with functions, as can be seen by the fact the Waml has a maximum arity it provides direct support for, expecting larger arities to fall back to arrays. These widespread techniques are already supported in a straightforward manner (i.e. no need for string mangling) with a nominal MVP.

> To that end, @rossberg, would you be able to list a few languages that you believe this proposal would enable to be separately compiled? If we can agree on what languages this proposal does support and what languages this proposal does not support, we will be in a good place.

I agree that answering this is the right next step for progressing the discussion.

rossberg:
Thanks @tlively for summarising.

> 1. Dynamic casts have to be implemented "manually" because distinct source types may have identical underlying structural types and the trick of placing them in the same `rec` group does not work across modules.
>    
>    * This is true of all the structural proposals so far. Anecdotally, the J2CL has been very happy to have this taken care of by nominal typing, but we don't have data on the actual size or performance impacts of manually implementing casting.

Yes. And this is true for all nominal proposals as well, as soon as the language has structural types or generics, i.e., every interesting language. It will be true even for J2CL once they get to actually implement casts on arrays, which, last time I asked, they hadn't yet.

The only way to avoid a manual implementation of language-level runtime types and instead fully piggy-back on Wasm RTTs is if Wasm's type system was expressive enough to embed the entire source language's dynamic type system. Obviously, that is unrealistic in general, and for any interesting language but Java it is immediately made impossible by the MVP's lack of generics.

Without Wasm generics, the closest you could use is a hybrid approach, with a first-level cast-check piggy-backed on Wasm RTTs and then a second-level manual check of generic parameters. But this second level still requires a full user-space implementation of the language's runtime type system, including runtime type canonicalisation! So in terms of effort, it does not buy anything (though the first level can be optimised by the engine, so might be a bit faster).

I assume this is how J2CL will deal with arrays (fortunately, Java's runtime types are so simplistic).

FWIW, I think I have pointed all this out on multiple occasions before, so it should not be a surprise.

> 2. Tuple types with subtyping form subtype semi-lattices rather than trees, requiring either the multiple supertypes extension or a separate subtype for each supertype with a coercion operator inserted where necessary.
>    
>    * This does seem complicated, but it's also worth noting that the currently implemented nominal system has this same problem and we have not discussed it at all or proposed any solutions. It would be good see a prototype of these potential solutions working, perhaps in Wob. By the way, what real-world languages have this kind of tuple subtyping?

Subtyping should be no harder than supporting generics over structural types in the MVP. Typically, when you have structural types, you need to employ a uniform representation for any of their type components that could be abstracted generically, since that may happen after the fact. And subtyping essentially just is a weaker form of polymorphism.

For example, in Wob, generics are restricted to reference types a priori, like in Java. Without Wasm-level generics, the way to achieve a uniform representation is to lower all reference components of a tuple to anyref. Hence, source-level subtyping on tuples has no impact on the representation level (and there is no subtyping on non-primitive types, which also is true for most languages).

As for mainstream examples of tuple subtyping, it's standard in modern OO languages:
```c#
// C#
class A {};
class B : A {};
(A,A) f((B,B) p) { return p; }
```
```scala
// Scala
class A {}
class B extends A {}
def f(p : (B, B)) : (A, A) = p
```
```typescript
// TypeScript
class A {}
class B extends A { x : number }
function f(p : [B,B]) : [A,A] { return p }
```
Scala e.g. represents tuples as an infinite family of generic classes Tuple2, Tuple3, ... Subtyping on them works because Scala supports variance annotations on generic parameters. This exemplifies my point that generics essentially introduce structural typing as well.


> 3. Exception tags are nominal, so they have to be coordinated in a nominal fashion, even when their payloads are structural types.
>    
>    * This has also been true of all the structural proposals so far. It's also true of the current nominal prototype, except that the nominal prototype would additionally coordinate everything else nominally as well. In practice the only way exception tags have been used so far is with a uniform representation for payloads, so this has not been a problem. Like for "manual" dynamic casting, we don't have data on the performance penalty we are paying by manually testing exception payload types.

Exceptions are completely different beasts. For one, they are values (value tags), not types. And as such, they have to be nominal for rather different reasons.

But yes, to implement a type-based exception-matching mechanism, one again needs a uniform representation.

> To that end, @rossberg, would you be able to list a few languages that you believe this proposal would enable to be separately compiled? If we can agree on what languages this proposal does support and what languages this proposal does not support, we will be in a good place.

There has been a lot of bogus claims and confusion here, so we need to clarify again what we mean by separate compilation. What I am referring to is the typical _incremental_ separate compilation model that many/most languages work with. Full separate compilation is way harder, not truly supported by most languages, and would require fundamental changes to Wasm's compilation model to be realistic. It thus is way out of scope. We also already agreed that we're not trying to solve the recursive linking problem.

With that context in mind, I cannot think of a language that would _not_ be supported by this proposal.

As I explained in my recent presentation, my overarching goal has been to avoid a notable regression over what you can do when targeting linear memory. That is, if you can lay out your data structures in linear memory and achieve some form of separate compilation with that, then you should be able to do the same when laying out the data in GC memory. It is _not_ a goal to introduce new magic mechanisms beyond that.

IOW, except actual memory management itself, everything that has to be done manually with linear memory still ought to (be able to) be done manually with GC memory. That reflects the nature of Wasm as a low-level assembly language.

RossTate:
> ```
> // C#
> class A {};
> class B : A {};
> (A,A) f((B,B) p) { return p; }
> ```

`(B,B)` is not a subtype of `(A,A)`. Neither of them are even reference types; they are value types. C# automatically inserts a coercion between the two types into the bytecode. But coercibility is weaker than subtyping, and as such the following program is rejected because they are not subtypes:
```
// C#
class A {};
class B : A {};
interface Covariant<out T> {}
Covariant<(A,A)> f(Covariant<(B,B)> p) { return p; } // type error: not subtypes
```

(TypeScript's type system is intentionally unsound and entirely erased.)

---

> There has been a lot of bogus claims and confusion here, so we need to clarify again what we mean by separate compilation. What I am referring to is the typical incremental separate compilation model that many/most languages work with.

[My comment](https://github.com/WebAssembly/gc/pull/243#issuecomment-930276173) already clarified that the issues were specifically with incremental compilation.

> With that context in mind, I cannot think of a language that would not be supported by this proposal.

This PR is substantially more complicated than what already exists. So it should be motivated by what it does *better* to merit that complexity. I gave a few examples of how it either doesn't help or even makes things worse (for specifically incremental compilation; there are also issues with module splitting and with separate compilation). The only languages where I have seen evidence that this proposal would make incremental compilation easier is OCaml and SML due to their (structural-)modules-are-heap-objects model. But that use case can also be addressed by a canonicalization feature *without* complicating the type system.

If there are advantages to this PR beyond making incremental compilation of OCaml and SML easier, one should be able to demonstrate them with concrete examples from major languages and implementation strategies. Providing such examples would really help advance the discussion.

tlively:
Yes, I believe we all have the same kind of incremental separate compilation in mind. To be even more concrete, we are talking about C++-style separate compilation where you have to recompile if your dependencies change their type layouts, for example by changing field order in a struct.

> With that context in mind, I cannot think of a language that would _not_ be supported by this proposal.

We already have @askeksa-google providing a concrete example of a language implementer wanting structural subtyping that this particular hybrid system does not support, so I am not as confident as you are about this. Given the available information so far, I suspect that we will need to see prototypes to be fully confident that this proposal works as intended, but it's worth seeing if we can make progress by discussing a few concrete examples first.

rossberg:
@tlively:
> > With that context in mind, I cannot think of a language that would _not_ be supported by this proposal.
> 
> We already have @askeksa-google providing a concrete example of a language implementer wanting structural subtyping that this particular hybrid system does not support

Hold on, @askeksa-google's observation is not about separate compilation, but about Dart using more expressive structural subtyping. And this would only be relevant as a counter example if Dart's source-level subtyping was mapped to Wasm-level representation subtyping. However, this won't be possible anyway: due to generics, Dart will already need to use a uniform representation AFAICS. So it's in the same camp as what I described for Wob above.

(AFAICS, the only way such a limitation could materialise right now would be for a language that has structural subtyping but no generics, and thus wants to avoid a uniform representation, which they couldn't. I'm not sure there is a major language in this category. If there was, they'd have to fall back to a uniform representation, i.e., casts. Of course, nominal typing wouldn't make that better.)

But again, this is before separate compilation even comes into play, and whatever choice is workable for a single module would be compatible with separate compilation, which is all I'm rooting for.

@RossTate:
> C# automatically inserts a coercion between the two types into the bytecode.

Fair enough, but I think we can agree that that is more a backwards-compat hack than anything else (tuples were only added late in C# 7.0, if I'm not mistaken).

> (TypeScript's type system is intentionally unsound and entirely erased.)

Yes, obviously, but doesn't affect @tlively's specific question that this was a reply to.

> [My comment](https://github.com/WebAssembly/gc/pull/243#issuecomment-930276173) already clarified that the issues were specifically with incremental compilation.

Pardon me, but the comment you are linking to does not clarify or explain anything, so there is little I can say to that.

As for casts in Wob, I said as much when I presented it: they are not implemented yet and they will need a user-space runtime type representation (I only need to find time to do that). And as I just explained again above, the same will be true for almost every language, including Java, under any Wasm type semantics. This part of the discussion is a red herring.

Finally, the semantics in this PR is not complicated implementation-wise. After (simple) type canonicalisation, its implementation in an engine is practically indistinguishable from one for nominal typing.

askeksa-google:
> Hold on, @askeksa-google's observation is not about separate compilation, but about Dart using more expressive structural subtyping. And this would only be relevant as a counter example if Dart's source-level subtyping was mapped to Wasm-level representation subtyping. However, this won't be possible anyway: due to generics, Dart will already need to use a uniform representation AFAICS. So it's in the same camp as what I described for Wob above.

To clarify, I certainly don't expect the WasmGC type system to be as expressible as the Dart one. That is not the issue.

Concretely, to handle subtyping of Dart function types (possibly with optional parameters) it looks like a good implementation strategy for function objects will involve vtables with subtyping relationships between them. For some of these, it can be useful to have the overlapping fields not have the exact same type, but rather change covariantly, as is allowed for immutable fields. It can also be useful to be able to assign function references to them with signatures that are subtypes of the field type.

The subtyping relationships between these Wasm function types could certainly be declared explicitly (provided that function types are allowed to declare more than one supertype), but it seems rather redundant to (have to) do so. The only reason I can see for having function supertypes be explicitly declared is if there was some use case for having some function types *not* be subtypes of each other, even though they would be structurally. But AFAICS, the only thing that can be achieved by this would be to make an otherwise sound module fail validation.

Exactly the same criticism applies to the nominal proposal.

RossTate:
> Fair enough, but I think we can agree that that is more a backwards-compat hack than anything else (tuples were only added late in C# 7.0, if I'm not mistaken).

As discussed in [this post](https://github.com/WebAssembly/gc/pull/243#issuecomment-930689793), C# intentionally does not support co/contravariant generic classes in order to support optimized representations. (Furthermore, `ValueTuple` is mutable, so that particular case would be unsound to be covariant.)

---

To investigate the claims here further, I decided to compare two approaches to a monomorphizing incremental C# compiler (that falls back to using a uniform representation where monomorphization is not possible). One generates (at compile time) a new module the first time a structural type like `(byte, float)` or `Tuple<byte,float>` is used, and then has subsequent uses reuse that module (I'll call this the memoizing approach). The other uses the proposed structural type system to avoid this (I'll call this the independent approach). I found the latter approach to be substantially worse. Here is why.

Consider what happens when a module has to compile the allocation `new Tuple<byte,float>(...)`. In addition to initializing the `byte` and the `float` field, the module needs to initialize the v-table for `Tuple<byte,float>`. With the memoizing approach, one can just call the exported constructor from the memoized module for `Tuple<byte,float>`, which will initialize the v-table to the v-table for `Tuple<byte,float>` stored in a global in the memoized module. For the independent approach, the module compiling `new Tuple<byte,float>` needs to create this v-table itself. This means it must also recreate the code for all the methods of `Tuple<byte,float>` in order to populate the v-table. Thus the cost of independence here is significant duplication of code.

Consider what happens when a module has to compile the surface-level cast `(Tuple<byte,float>)`. In the memoizing approach, we can implement this cast in constant time because the memoizing module can ensure that there is a single casting identifier used by all `Tuple<byte,float>` instances. That is, memoizing also provides custom canonicalization. (With nominal types, they could even piggy back off of the built-in casting mechanism.) In the independent approach, we have no canonicalization, so surface-level casts to even class types can only be implemented in linear time.

Both approaches have problems handling the variance of delegates and interfaces for the reasons discussed above (namely function subtyping as well as lattices). In both cases you have to represent values of co/contravariant type variables in the uniform representation in order to use typed function references. But with the memoizing approach you can use call tags in order to use specialized representations. That is, if you call a method `bar` of an interface `Foo<float,byte>`, you can use a call tag for that method provided by the memoized module for which the `float`/`byte` values are passed in their specialized (i.e. unboxed) forms. Any monomorphized classes implementing the interface can directly handle the specialized call tag. And for other cases, the default behavior of the call tag can simply box the values into their uniform representation and pass them to the same receiver using the general-purpose call tag instead (which all implementing classes are guaranteed to handle).

So, for the same practical reasons that most languages only support tuples up to a maximum arity, I believe the memoization approach would be the more practical option for supporting structural features of most languages. OCaml/SML module-instances-as-heap-objects seem to be outliers among structural languages, and even they could still use the memoizing approach (though I agree that the independent approach is the nicer fit for these two languages).

rossberg:
@askeksa-google:
> The subtyping relationships between these Wasm function types could certainly be declared explicitly (provided that function types are allowed to declare more than one supertype), but it seems rather redundant to (have to) do so. The only reason I can see for having function supertypes be explicitly declared is if there was some use case for having some function types _not_ be subtypes of each other, even though they would be structurally. But AFAICS, the only thing that can be achieved by this would be to make an otherwise sound module fail validation.

Yes, I agree, but at the risk of restating the obvious: none of what you say is specific to function types, it applies equally to structs. Consequently, the purpose of this change to the proposal isn't to enable new things for producers, it merely is to tame subtyping and recursion. And this combination is no easier for function types, where it comes up as well. So unfortunately, something has to give.

FWIW, there is one upside to making function subtyping explicit: we could readily allow call_indirect to coherently check the callee type modulo subtyping, without unpredictable performance costs.

askeksa-google:
> Yes, I agree, but at the risk of restating the obvious: none of what you say is specific to function types, it applies equally to structs.

No, the key difference is that structs support runtime subtype tests, whereas functions do not. Compact and efficient runtime tests usually require (statically or dynamically) declared single inheritance.

> Consequently, the purpose of this change to the proposal isn't to enable new things for producers, it merely is to tame subtyping and recursion. And this combination is no easier for function types, where it comes up as well. So unfortunately, something has to give.

Isn't the purpose rather to tame *type equivalence*? CMIIW, but it seems that with equivalence in place, subtyping is straightforward (recursive traversal with memoization). I looked at the V8 code for subtyping of (possibly) recursive function types, and it was not complicated.

> FWIW, there is one upside to making function subtyping explicit: we could readily allow call_indirect to coherently check the callee type modulo subtyping, without unpredictable performance costs.

We could, yes. I am not so sure it would be a good idea, though, for a few reasons:
- It would limit function types to a single supertype, which could lead to some unwieldy situations.
- `call_indirect` is slow enough as it is. Adding another full subtype check on top could severely affect performance, unless it enables savings somewhere else.
- The main issue with the exact signature requirement of `call_indirect` is that it requires all functions that can be called from the same call site to type their parameters as the LUB of the "true" types for the particular parameter across the functions. This would still be the case if `call_indirect` allowed subtyping. This would just allow some of the functions to have even less precise types on some parameters (or more precise return types, of course). This can only be useful if the function is also called from somewhere else. Is there a particular source language situation where this would be useful?

rossberg:
> > Yes, I agree, but at the risk of restating the obvious: none of what you say is specific to function types, it applies equally to structs.
> 
> No, the key difference is that structs support runtime subtype tests, whereas functions do not.

In the proposal as is, you can cast functions just like structs. There just wasn't a way before to assign a non-canonical RTT to a function, but that possibility is [listed in the Post-MVP doc](https://github.com/WebAssembly/gc/blob/master/proposals/gc/Post-MVP.md#custom-function-RTTs). After this change, that wouldn't even be needed.

> Isn't the purpose rather to tame _type equivalence_? CMIIW, but it seems that with equivalence in place, subtyping is straightforward (recursive traversal with memoization). I looked at the V8 code for subtyping of (possibly) recursive function types, and it was not complicated.

Both, actually. Structural subtyping is not algorithmically complicated, but algorithmically complex. ;) Worse so than equivalence. Even with memoisation and the simple type structure we currently have, it's quadratic in the worst case. And it can get way worse with generics.

> > FWIW, there is one upside to making function subtyping explicit: we could readily allow call_indirect to coherently check the callee type modulo subtyping, without unpredictable performance costs.
> 
> We could, yes. I am not so sure it would be a good idea, though, for a few reasons:
> * It would limit function types to a single supertype, which could lead to some unwieldy situations.
> * `call_indirect` is slow enough as it is. Adding another full subtype check on top could severely affect performance, unless it enables savings somewhere else.

Not necessarily. It would be possible to decorate and constrain subtyping such that the compiler would always be able to tell statically whether the subtype hierarchy is empty, linear or neither. Under such a condition, it would remain pay-as-you-go. But yes, this may not be worth the complexity.

(However, without it we'll need to fix the problem that introducing subtyping on functions without explicit rtt.sub would break important type system properties if call_indirect was unmodified, such as the ability to refine static types without affecting execution. There are ways to do that, but they're not necessarily nicer. It's harder if subtypes do not need to be declared.)

RossTate:
For what it's worth, there is a static solution to the method-signature-refinement problem using an extension to #245, and there is a dynamic solution to the same problem using WebAssembly/call-tags#3 (that our own experiments suggest performs quite well). The dynamic solution can also accommodate changes in representation (e.g. due to boxing). Both of those solutions are compatible with incremental compilation, whereas the LUB solution using this PR is not. I am happy to go into more depth on those solutions, but my impression is that people are reluctant to expand upon the nominal approach while this PR is still in consideration. So I would like to see this PR come to some resolution soon.

tlively:
Right now I don't think we have sufficient concrete evidence to conclude that this proposal is or is not sufficiently more expressive than nominal types to enable additional real-world use cases. It's certainly less expressive for structural subtyping than the combined nominal and equirecursive approach, which we have not finished evaluating, and it's unclear whether that expressiveness gap matters in practice.

To get more evidence, it would be best to develop a prototype. I would be happy to prototype this system in Binaryen if we had any producers interested in prototyping separate compilation and dynamic linking for their language, but we don't have any such producers right now. Even if we had producers interested in prototyping incremental compilation (without dynamic linking), I would recommend that they first try to use LLVM's object file format and linker, for which the type system would not matter.

Assuming we don't find a "real-world" producer interested in separate compilation and dynamic linking in the near future, I think the best we could do to move this conversation forward would be to make Wob/Waml sufficiently "real-world" that they could give us confidence that this proposal would work for separately/incremental compilation of real-world languages (and also work better than just nominal types or nominal+equirecursive types).

tlively:
I have a PR up for isorecursive type canonicalization in Binaryen here: https://github.com/WebAssembly/binaryen/pull/4481.

One question that was raised when I was discussing the implementation with some other folks was whether types not within recursion groups were meant to be equivalent to types within recursion groups of size one. Specifically, is it valid for a type not within a recursion group to contain itself as a child? In the Binaryen implementation, I rely heavily on types without recursion groups being syntactic sugar for types with arity-1 recursion groups, which lets me uniformly treat all defined types as having recursion groups. @rossberg, is that the intended semantics, or are types without recursion groups meant to actually not have recursion groups? If that's not the intended semantics, would we be able to make it the intended semantics?

Another note is that I was thinking about implementing the algorithm for minimizing recursion groups and talking it over with @kripken, and I realized that if the minimization preserves separate type identities for structurally-identical types that are originally in the same recursion group, then the result is not suitable for the kind of ad-hoc after-the-fact linkage based on complete knowledge of a single type that I had expected the minimization algorithm to support. The problem is that separate-but-identical types `A` and `A'` have to be in the same recursion group after minimization to maintain their separate identities, so another module trying to create a minimal recursion group for `A` will not get the same group unless it accounts for the potentially unrelated type `A'` as well. I had been hoping that recursion group minimization would be a safe transformation that would allow for ad-hoc linking for free, but it turns out it can be safe and not quite get you to ad-hoc linking or it can be unsafe (by merging identical types) and get you to ad-hoc linking.

askeksa-google:
Indeed, in order to produce deterministic, canonical recursion groups that can be used for linking, the minimization would need to do essentially equi-recursive canonicalization and then partition the result into recursion groups.

What do you mean that merging these types is unsafe? Is it any more unsafe than merging types in different recursion groups, which the minimization would do in any case?

As I see it, there are two useful modes that tools like Binaryen can operate in with regards to type minimization:
1. Perform equi-recursive normalization, then create recursion groups in a canonical way (i.e. a group containing the same types, depending on identical other groups, always end up in the same order within the group). This will work with linking if the same canonicalization is performed on all linked modules.
2. Preserve all recursion groups, including the ordering of types within them. Identical groups can be merged. This will work with linking even if some modules are transformed and some are not.

tlively:
> What do you mean that merging these types is unsafe? Is it any more unsafe than merging types in different recursion groups, which the minimization would do in any case?

Yes, there's a difference. Merging two definitions of the same type (i.e. deduplicating a recursion group) is always safe because the original two definitions were not observably different. Merging duplicated types in a single recursion group is observable, though, because a `br_on_cast` testing whether a value of one type also has the other type will take the branch after merging the types but not before. To do this merging safely, producers will have to make sure their programs do not observe the differences between the original types, which is not something Binaryen can prove in general.

Doing equirecursive minimization is similarly unsafe because a program could observe the differences between two types that are isorecursively distinct but equirecursively equivalent. This would still be a nice code size optimization for Binaryen to offer, but producers would probably have to opt in since its safety relies on properties Binaryen can't prove.

rossberg:
@tlively: Re non-recursive types: It's a good question, and one that I have wondered about myself.  The current implementation distinguishes non-recursive types from singleton groups, mostly to be conservative. But I could not think of any practical use case for distinguishing them. So I'd be fine with identifying them.

Not sure I fully understand the scenario you are describing in your second comment. It is correct that splitting or modifying recursion groups is not a semantics-preserving transformation, since it changes type equivalences. So a tool like Binaryen must not do it by default. That has to be done (or at least requested) by producers. Binaryen can still dedupe equivalent groups, though.

(It's safe to transform a recursion group only if the module does not depend on type equivalence for any of the types contained, nor does it im/export anything involving these types. But the precise analysis would require some care. Basically, you'd have to prove that none of the types flows into a cast or call_indirect, at least not with different but prior equivalent type definitions.)

tlively:
Thanks, I can't think of a use case for separate non-recursive types either.

The scenario I was describing is when a frontend wants to delegate the work of figuring out the canonicalized group contents to Binaryen, with the goal of allowing ad-hoc linking with the output. In that situation, the groups in the input to Binaryen are arbitrary and not meant for public consumption, so it is assumed that Binaryen will change their identity. Beyond that desired change of type identity, I was hoping that the transformation would be safe in all cases, but it turns out to be unsafe (or alternatively not as useful) if the module observes the difference between identical types in the same rec group.

askeksa-google:
I think it's safe to assume that if a producer wants Binaryen to perform type minimization, then it does not depend on distinguishing the identities of structurally identical types. Otherwise, as you say, there isn't really much that Binaryen can do.

So the flag controlling this behavior in Binaryen will be a signal from the producer that it does not need this distinction, which then enables Binaryen to perform the minimization / canonicalization.

If the producer needs to distinguish between certain types, it will be its responsibility to create recursion groups that uphold that distinction.

kripken:
To make sure I understand, is it correct that we are expecting two workflows here? Like this:

1. The producer **expects** to link wasm modules at runtime. It emits proper recursion groups for that. An optimizer like Binaryen must not modify those recursion groups in any way (except, I guess, for removing an entire recursion group when possible).
2. The producer **does not expect** to link modules at runtime. An optimizer like Binaryen can freely optimize types inside the module (removing types, removing fields, etc.). In this case, it would be best to emit a single big rec group, both because it saves some space in the type section, and also because that defines types in the most precise way possible (there is no automatic merging of types that would limit toolchain or VM optimizations).

And optimizers like Binaryen will need a flag to distinguish between these two cases. An open question is what the default should be for such a flag.

Does all that make sense? Or is there a third use case, or is something wrong in one of the use cases?

askeksa-google:
There is a third use case, the first in https://github.com/WebAssembly/gc/pull/243#issuecomment-1024188245: The producer expects to link its Wasm modules, but it will always optimize its modules with Binaryen (or a similar tool, but crucially the same for all modules). Here, the optimizer is free to change the recursion groups as long as so does it in a deterministic manner (e.g. equi-recursive canonicalization) to make the groups consistent across the modules.

kripken:
@askeksa-google I see, yes, I guess as long as the optimizations are "deterministic" (do not change depending on the code in each module) then that could work.

But how much benefit do you expect there? My intuition is that the large improvements come from optimizations that do depend on the code, like removing unused types and fields, making fields immutable, etc.

tlively:
I measured the type validation and canonicalization time for a real-world j2wasm module across three type systems, both before and after applying Binaryen optimizations at -O2. For each type system, I measured the total validation and canonicalization time as well as times for two separate components of the work. The first component involves validating the new heap type definitions and deduplicating their definitions and the second component involves moving the new heap types and copying the new value types to the global type stores. The implementation of the second part is the same for all three systems, but there's probably some room to optimize it for the nominal and isorecursive systems. The modules have single large recursion groups in the isorecursive case. I am also interested in measuring the case of minimal recursion groups, but don't have the modules in that format yet.

The unoptimized module has 88442 type definitions and is about 93 MB, including a name section. Validating and building these types takes:

 - Nominal: 402 ms (19 ms + 381 ms)
 - Isorecursive: 451 ms (69 ms + 382 ms)
 - Equirecursive: 14855 ms (493 ms + 14361 ms)

The optimized module has 24843 type definitions and is about 10 MB without a name section. Validating and building these types takes:

  Nominal: 89 ms (4 ms + 84 ms)
  Isorecursive: 104 ms (19 ms + 85 ms)
  Equirecursive: 4282 ms (139 ms + 4143 ms)

My takeaway is that the extra overhead of canonicalizing isorecursive types over nominal types is probably not an issue, especially considering that the canonicalization algorithm is linear.

tlively:
I also measured an upper bound on the code size cost of the additional constraints on the order of types in the type section. In nominal and equirecursive modes, Binaryen orders types by decreasing use count so that frequently used types have smaller LEB-encoded indices. I measured the same optimized module from the previous comment with the optimal default ordering, an ordering that respects the isorecursive constraints on the ordering of subtypes but otherwise tries to be optimal, and the reverse of the optimal ordering. Again, the isorecursive case uses a single large recursion group, so this is the least constrained case for the isorecursive system, but even in the worst case the isorecursive system can be no worse than the pessimal ordering.

Optimal nominal ordering (baseline):

 - binary size: 10104049 bytes
 - gzipped binary size: 2402160 bytes

Isorecursive ordering:

 - binary size: 10106225 bytes (+0.02%)
 - gzipped binary size: 2400039 bytes (-0.09%)

Pessimal nominal ordering:

 - binary size: 11205902 bytes (+10.9%)
 - gzipped binary size: 2438062 (+1.5%)

So I don't think binary size will be a problem for the isorecursive type system either.

askeksa-google:
@kripken I don't see optimization as the main goal in this scenario. Rather, it allows a producer to not perform any type canonicalization itself, leaving it all to e.g. Binaryen, while still being able to link its modules at runtime.

kripken:
@askeksa-google Ok, that's where we were seeing things differently then. I was only thinking about optimizations, where I think a single flag is enough. Yes, makes sense that if there are other non-optimization stuff we should do in Binaryen then those could be additional flags as needed.

proposals/gc/MVP.md
```diff
@@ -75,15 +74,24 @@ New abbreviations are introduced for reference types in binary and text format,
...
 
 
 #### Type Definitions
 
-* `deftype` is a new category of types that generalises the existing type definitions in the type section
-  - `deftype ::= <functype> | <structtype> | <arraytype>`
+* `deftype` is the syntax for an entry in the type section, generalising the existing syntax
+  - `deftype ::= <subtype> | rec <subtype>*`
   - `module ::= {..., types vec(<deftype>)}`
+  - a `rec` definition defines a group of mutually recursive types that can refer to each other; it thereby defines several type indices at a time
```
tlively, line 86: Do we intend to support groups of size 0? It would slightly simplify the Binaryen implementation if we did not.

rossberg:
@tlively, thanks for these numbers. That's encouraging!

FWIW, if necessary, I wouldn't mind imposing a different constraint to ensure acyclicity in the subtype relation. The ordering restriction merely seemed the simplest and cheapest to check.

proposals/gc/MVP.md
rossberg: Ah, good question. Preferably, I'd go with what we do everywhere else in Wasm, which is avoiding special rules for edge cases. How much of a simplification would it be to rule it out?

tlively: That's fine. It will mean adding the ability to shrink a `TypeBuilder` in Binaryen, but that's not too complicated.

proposals/gc/MVP.md
```diff
@@ -75,15 +74,24 @@ New abbreviations are introduced for reference types in binary and text format,
...
+
+* `subtype` is a new category of type defining a single type, as a subtype of possible other types
+  - `subtype ::= sub <typeidx>* <strtype>`
+  - the preexisting syntax with no `sub` clause is redefined to be a shorthand for a `sub` clause with empty `typeidx` list: `<strtype> == sub () <strtype>`
+  - Note: This allows multiple supertypes. For the MVP, it could be restricted to at most one supertype.
```
tlively, line 91: Let's change the language from "could be restricted" to "is restricted."

proposals/gc/MVP.md
```diff
@@ -97,20 +105,187 @@ New abbreviations are introduced for reference types in binary and text format,
...
+
+For the purpose of defining recursive type equivalence, type indices are extended with a special form that distinguishes regular from recursive type uses.
+
+* `rec.<i>` is a new form of type index
+  - `typeidx ::= ... | rec.<i>`
+
+This form is only used during equivalence checking, to identify and represent "back edges" inside a recursive type. It is merely a technical device for formulating the rules and cannot appear in source code. It is introduced by the following auxiliary meta-function:
+
+* Rolling a context type produces an _iso-recursive_ representation of its underlying recursion group
+  - `tie($t)                    = tie_$t(<ctxtype>)`  iff `$t = <ctxtype>`
```
tlively, line 204: Should this be "iff `C($t) = <ctxtype>`" like in the `unroll` and `expand` definitions?

proposals/gc/MVP.md
```diff
@@ -97,20 +105,187 @@ New abbreviations are introduced for reference types in binary and text format,
...
+Note: In type-theoretic terms, these are higher-kinded iso-recursive types:
+```
+tie($t1) ~ (mu a. <(struct (field i32 (ref a.1))), (struct i64 (field (ref a.0)))>).0
+tie($t2) ~ (mu a. <(struct (field i32 (ref a.1))), (struct i64 (field (ref a.0)))>).1
+```
+where `<...>` denotes a type tuple. However, in our case, a single syntactic type variable `rec` is enough for all types, because recursive types cannot nest by construction.
+
+Note 2: This semantics implies that type equivalence checks can be implemented in constant-time by representing all types as trees in tied form and canonicalising them bottom-up in linear time upfront.
+
+Note 3: It's worth noting that the only observable difference to a nominal type system is the equivalence rule on (non-recursive) type indices: instead of looking at their definitions, a nominal system would require `$t = $t'` syntactically (at least as long as we ignore things like checking imports, where type indices become meaningless).
```
tlively, line 270: Other differences are the presence of recursion groups and the restrictions on which type indices can be used in the definition of a type. Maybe it would be more useful to note here that having a single large recursion group essentially gives you a nominal system?

proposals/gc/MVP.md
```diff
@@ -97,20 +105,187 @@ New abbreviations are introduced for reference types in binary and text format,
...
 #### Subtyping
 
-Greatest fixpoint (co-inductive interpretation) of the given rules (implying reflexivity and transitivity).
+##### Type Indices
+
+In the [existing rules](https://github.com/WebAssembly/function-references/proposals/function-references/Overview.md#subtyping), subtyping on type indices required equivalence. Now it can take declared supertypes into account.
+
+* Type indices are subtypes if they either define [equivalent](#type-equivalence) types or a suitable (direct or indirect) subtype relation has been declared
+  - `$t <: $t'`
+    - if `$t = <ctxtype>` and `$t' = <ctxtype'>` and `<ctxtype> == <ctxtype'>`
```
tlively, line 281: Should these be `C($t)` and `C($t')`?

proposals/gc/MVP.md
```diff
@@ -123,20 +298,16 @@ In addition to the [existing rules](https://github.com/WebAssembly/function-refe
   - `i31 <: eq`
 
 * Any concrete type is a subtype of either `data` or `func`
-  - `(type $t) <: data`
+  - `$t <: data`
      - if `$t = <structtype>` or `$t = <arraytype>`
-     - or `$t = type ht` and `rt <: data` (imports)
-  - `(type $t) <: func`
+     - or `$t = type ht` and `ht <: data` (imports)
```
tlively, line 303: Are these lines referring to type imports? I know they were already in here, but let's remove them since we don't plan to have type imports in the MVP.

proposals/gc/MVP.md
rossberg: The comment was referring to the difference in the rules. I reworded it to make that more explicit, and also added a sentence describing the practical consequence you mention.

proposals/gc/MVP.md
rossberg: See above.

proposals/gc/MVP.md
rossberg: Right, removed dependence on type imports.

proposals/gc/MVP.md
```diff
@@ -176,43 +341,38 @@ Note: In the future, this hierarchy could be refined to distinguish compound dat
...
+Note: RTT values correspond to type descriptors or "shape" objects as they exist in various engines. RTT equality can be implemented as a single pointer test by memoising RTT values. More interestingly, runtime casts along the hierachy encoded in these values can be implemented in an engine efficiently by using well-known techniques such as including a vector of its (direct and indirect) super-RTTs in each RTT value (with itself as the last entry). A subtype check between two RTT values can be implemented as follows using such a representation. Assume RTT value v1 has static type `(rtt $t1)` and v2 has type `(rtt $t2)`. Let `n1` and `n2` be the lenghts of the respective supertype vectors. To check whether v1 denotes a subtype RTT of v2, first verify that `n1 >= n2` -- if both `n1` and `n2` are known statically, this can be performed at compile time; if either is not statically known, it has to be read from the respective RTT value dynamically, and `n1 >= n2` becomes a dynamic check. Then compare v2 to the n2-th entry in v1's supertype vector. If they are equal, v1 is a subtype RTT.
```
tlively, line 361: A couple spelling fixes and remove the mention of n1 and n2 not being statically known.
```suggestion
Note: RTT values correspond to type descriptors or "shape" objects as they exist in various engines. RTT equality can be implemented as a single pointer test by memoising RTT values. More interestingly, runtime casts along the hierarchy encoded in these values can be implemented in an engine efficiently by using well-known techniques such as including a vector of its (direct and indirect) super-RTTs in each RTT value (with itself as the last entry). A subtype check between two RTT values can be implemented as follows using such a representation. Assume RTT value v1 has static type `(rtt $t1)` and v2 has type `(rtt $t2)`. Let `n1` and `n2` be the lengths of the respective supertype vectors. To check whether v1 denotes a subtype RTT of v2, first verify that `n1 >= n2` -- both `n1` and `n2` are known statically, so this can be performed at compile time. Then compare v2 to the n2-th entry in v1's supertype vector. If they are equal, v1 is a subtype RTT.
```

tlively (approved):
LGTM!

proposals/gc/MVP.md
rossberg: Ah, but for casts the type `$t1` itself usually is not statically known (only a supertype of it), so neither is `n1`. I pushed a tweak that points that out.

Issue #259, opened 2021-12-08
Supertype of all arrays


askeksa-google:
In `dart2wasm`, there are two different representations for strings - `OneBytetring` (containing Latin-1) and `TwoByteString` (containing WTF-16). Each is a wrapper struct containing some fields present in all Dart objects (class id, identity hash) plus a Wasm array. The element type of the array is `i8` and `i16`, respectively.

The `.length` getter on a string is a virtual call, since it will need to dispatch to specific code for reading the length of the array of `i8` or `i16`. This seems a bit silly, since the underlying engine likely has a partly uniform representation of arrays where the length could be accessed in the same place regardless of the element type.

This gave me an idea: we could have an `arrayref` type which is a subtype of `dataref` and a supertype of all arrays. The only operation permitted on an `arrayref` (on top of the ones for `dataref`) is `array.len`.

With this, the common string supertype of the two string types could declare an immutable field of type `arrayref` for the contents array, which would then be specialized in each subclass. The `.length` getter could simply execute an `array.len` on that field.

WDYT?

titzer:
This sounds like a good idea to me. In Java class files, the `arraylength` bytecode will work on any type of array.

rossberg:
This makes sense to me. But I assume we'd also need accompanying instructions like `ref.is_array` and `ref.as_array` to actually get values of this type.

askeksa-google:
The `arrayref` type is useful even without these instructions (via implicit upcasts and field covariance), but it makes sense to add them for consistency.

manoskouk:
To make this work we will have to remove the immediate array index from `array.len`.

manoskouk:
Update: We just implemented `arrayref` in V8 including `ref.is_array`, `ref.as_array`, `br_on_array` and `br_on_non_array`. The opcodes and details can be found [here](https://docs.google.com/document/d/1DklC3qVuOdLHSXB5UXghM_syCh-4cMinQ50ICiXnK3Q/). Notably, we now ignore the type index immediate for `array.len` so that it can work on values of type `anyref`. For backwards compatibility, we did not drop the immediate yet.

rossberg:
Thanks for taking the lead @manoskouk. I backported your implementation to this repo, please see #264.

tlively:
I'll close this since the new type has been merged into the explainer and no one has argued against it.

PR #264, opened 2022-01-18
Add arrayref and resp instructions

rossberg:
Addresses #259.

Follows the lead of the V8 implementation and implements the same opcodes as specified [here](https://docs.google.com/document/d/1DklC3qVuOdLHSXB5UXghM_syCh-4cMinQ50ICiXnK3Q/).

Includes MVP spec, tests, and interpreter implementation.

@manoskouk, PTAL.


interpreter/binary/decode.ml
```diff
@@ -188,6 +189,7 @@ let ref_type s =
   | -0x17 -> let n = vu32 s in (NonNullable, RttHeapType (var_type s, Some n))
   | -0x18 -> (NonNullable, RttHeapType (var_type s, None))
   | -0x19 -> (Nullable, DataHeapType)
+  | -0x20 -> (Nullable, ArrayHeapType)
```
manoskouk, line 192: There is a difference with the V8 prototype, as well as MVP.md: there, arrayref, i31ref and dataref are used as shorthands for the non-nullable versions of the types.

proposals/gc/MVP.md
```diff
@@ -522,7 +538,7 @@ The opcode for heap types is encoded as an `s33`.
 | 0xfb14 | `array.get_s $t` | `$t : typeidx` |
 | 0xfb15 | `array.get_u $t` | `$t : typeidx` |
 | 0xfb16 | `array.set $t` | `$t : typeidx` |
-| 0xfb17 | `array.len $t` | `$t : typeidx` |
+| 0xfb17 | `array.len` | `u32` (TODO: remove, was typeidx) |
```
manoskouk, line 541: Shouldn't this be
 | 0xfb17 | `array.len $t` | `$t : u32` |

manoskouk (approved):
LGTM with a couple of comments.

interpreter/binary/decode.ml
rossberg: Yeah, for now I went with what V8 does (required changing some tests), but I opened #266 to perhaps fix this.

proposals/gc/MVP.md
rossberg: Well, the index has been removed from the abstract syntax, so the parameter is merely a detail of the binary format. Changed to `_ : u32`.

PR #270, opened 2022-02-03
Change arrayref opcode

### LINKED ISSUE #269: Type code clash for `array` and function type constructors
#264 added the arrayref type, following the V8 implementation. However, this reuses the type code -0x20 (0x66) for the array heap type, which is already used for function types. In the presence of type exports, we will likely need to allow abstract heap types like `array` in type definitions, so this clash is a problem.

I suggest changing to -0x1a instead, which was skipped anyway.

rossberg:
Changes opcode from -0x20 to -0x1a. Fixes #269.


jakobkummerow (approved):
LGTM

PR #273, opened 2022-02-09
Structref

rossberg:
Since this came up in yesterday's meeting, this PR adds a structref type and corresponding instructions, for symmetry with arrayref and funcref.


jakobkummerow:
Does this have any other reason going for it than "for symmetry"?

rossberg:
None right now. It would fall out rather naturally from #274, so I'm happy to defer.

jakobkummerow:
Now that we got `structref` via #325, should probably close this PR to not create confusion.

Issue #266, opened 2022-01-19
Reconcile reftype short hands wrt nullability


rossberg:
Currently, i31ref, dataref, and arrayref expand to non-null, while anyref, eqref, funcref, and externref are null. Should probably be more principled.

tlively:
Yes, please. I'm always forgetting which ones are null and which are not. I suggest that we make them all nullable because writing `anyref` instead of `(ref any)` is less of an improvement than writing it instead of `(ref null any)`.

rossberg:
Makes sense to me. Other opinions?

manoskouk:
We are already committed to `funcref` and `externref` being nullable, therefore if we want a uniform treatment, nullable is the only choice we have without making breaking changes. `i31ref` is somewhat counterintuitive as nullable in my opinion, although that might not be relevant depending on the resolution of #130.

titzer:
I agree with making all the shorthands nullable.

jakobkummerow:
(Moving my comment from the PR)

I think it would be nice to get some data: take as many GC-using Wasm modules as you can find, and count how often they use `(ref X)` and `(ref null X)` for all relevant values of X. Then we can define the meaning of the shorthands such that they actually save a maximum number of bytes. If we find that they don't save many bytes at all, then we might as well drop them entirely, as replacing a two-byte `(ref X)` or `(ref null X)` with a one-byte `xref` is pretty much their only purpose.

Note that I'm not opposed to flipping every single default if need be. I'm just saying it'd be good to have a solid reason for doing so. With Wasm being a compilation target, convenience, aesthetics, and symmetry/regularity/uniformity matter far less than they do in human-written languages, and hence are not convincing arguments for making design decisions.

tlively:
FWIW, I've mostly been thinking of this as a text format change and I only recently connected the dots and realized there would be a binary encoding change as well.

I just measured the effect of never using the binary shorthands on the 10MB optimized J2CL module and it only increased code size by 13 bytes. I assume that's not representative of all use cases, but after seeing that I would certainly be in favor of making the text shorthands nullable and removing the binary shorthands that haven't already been standardized. Or removing the new text shorthands as well.

rossberg:
Interesting. Does J2CL use non-concrete types or i31 much, though?

I did the same experiments with removing all new binary shorthands and saw a +0.34% size increase on the total size of the Waml test suite modules and +0.64% for the Wob test suite.

[...]

PR #277, opened 2022-02-18
Make nullable the default for reference type shorthands

rossberg:
Addresses #266.


rossberg:
@jakobkummerow, #266 would be the right place to discuss this. So far, all comments were in favour of this change.

titzer (approved):

Issue #260, opened 2021-12-08
Initializing array from data or element segment


askeksa-google:
In various discussions, we have touched upon the subject of having a version of the `array.init` instruction which initializes the array from a data or element segment rather than from the stack. Most recently, @rossberg suggested some variations of the operation in https://github.com/WebAssembly/gc/issues/250#issuecomment-943276626.

Of the various proposed instructions for this purpose, I think the ones that create a new array are the most important, as these provide new capabilities for global initializers. These would be (staying with the `init` nomenclature):
```
array.init_from_data $a (data $d) : [i32 i32 (rtt $a)] -> [(ref $a)]
array.init_from_elem $a (elem $e) : [i32 i32 (rtt $a)] -> [(ref $a)]
```
each taking an offset (byte resp. element) and the array length from the stack.

I did an experiment in `dart2wasm`, replacing the use of `array.init` for string constants with `array.init_from_data` as specified above.

For a module with 34k of string data, it reduced the module size by 48k uncompressed. Compressed (zopfli), the size reduction was less dramatic: 300 bytes if the instruction takes the offset first, then the length, or 1200 bytes if it takes the length followed by the offset.

But this is not merely a matter of module size reduction. Due to the limitation of `array.init` that there is a maximum length it can accept, these instructions are essential to enable arrays of any length to be initialized in global initializers. This is important for the implementation of Dart constant expressions, for instance.

tlively:
SGTM, and I think the names are appropriate as well. The `memory.init` instruction uses [dest_offset, source_offset, length] order for its operands, so it would be good to use [source_offset, length] in these instructions. Interesting that the compression was better the other way around, though.

askeksa-google:
One thing to keep in mind is the interaction with `data.drop` and `elem.drop`.

Ordinarily, constant expressions are independent of all side effects, which means an engine is free to postpone initialization of a global until e.g. the first time a function using the global is compiled. However, if we specify that these instructions will trap when referring to a dropped data or element segment (which would be a natural thing to do), then such postponing becomes observable.

To restore the illusion that global initialization happens fully up front, the engine would need to perform at least a quick pass over the global section to identify references to data and element segments. It can then instantiate those arrays up front or keep the segments alive internally until all uses in initializers have been executed.

rossberg:
[...]

PR #278, opened 2022-02-18
Add bulk array creation instructions

rossberg:
Addresses #260, following the V8 prototype spec, by adding:

* `array.new_data`
* `array.new_elem`
* `array.new_fixed`

Also adds additional bulk operators to Post-MVP doc.

Instruction names differ from V8 doc to clarify where allocation happens and for consistency with memory and table instructions.

Moreover, make all `*.new` instructions constant, except `array.new_elem`.

### DIFF
--- a/proposals/gc/MVP.md
+++ b/proposals/gc/MVP.md
@@ -460,12 +460,34 @@ This can compile to machine code that (1) reads the RTT from `$x`, (2) checks th
 
 * `array.new_with_rtt <typeidx>` allocates an array with RTT information determining its [runtime type](#values)
   - `array.new_with_rtt $t : [t' i32 (rtt $t)] -> [(ref $t)]`
-    - iff `expand($t) = array (var t')`
+    - iff `expand($t) = array (mut t')`
 
 * `array.new_default_with_rtt <typeidx>` allocates an array and initialises its fields with the default value
   - `array.new_default_with_rtt $t : [i32 (rtt $t)] -> [(ref $t)]`
-    - iff `expand($t) = array (var t')`
+    - iff `expand($t) = array (mut t')`
     - and `t'` is defaultable
+
+* `array.new_fixed <typeidx> <N>` allocates an array of fixed size and initialises it from operands
+  - `array.new_fixed $t N : [t^N (rtt $t)] -> [(ref $t)]`
+    - iff `expand($t) = array (mut t')`
+
+* `array.new_data <typeidx> <dataidx>` allocates an array and initialises it from a data segment
+  - `array.new_data $t $d : [i32 i32 (rtt $t)] -> [(ref $t)]`
+    - iff `expand($t) = array (mut t')`
+    - and `t'` is numeric or packed numeric
+    - and `$d` is a defined data segment
+  - the 1st operand is the `offset` into the segment
+  - the 2nd operand is the `size` of the array
+  - traps if `offset + |t'|*size > len($d)`
+
+* `array.new_elem <typeidx> <elemidx>` allocates an array and initialises it from an element segment
+  - `array.new_elem $t $e : [i32 i32 (rtt $t)] -> [(ref $t)]`
+    - iff `expand($t) = array (mut t')`
+    - and `t'` is a reference type
+    - and `$e` is a defined element segment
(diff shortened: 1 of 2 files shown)


proposals/gc/MVP.md
```diff
@@ -737,6 +759,9 @@ The opcode for heap types is encoded as an `s33`.
 | 0xfb15 | `array.get_u $t` | `$t : typeidx` |
 | 0xfb16 | `array.set $t` | `$t : typeidx` |
 | 0xfb17 | `array.len` | `_ : u32` (TODO: remove, was typeidx) |
+| 0xfb19 | `array.new_fixed $t N` | `$t : typeidx`, `N : u32` |
+| 0xfb1d | `array.new_data $t $d` | `$t : typeidx`, `$d : dataidx` |
+| 0xfb1e | `array.new_elem $t $e` | `$t : typeidx`, `$e : elemidx` |
```
tlively, line 764: @manoskouk, does the current `array.init_from_data` instruction in V8 work for both data and element segments or just for data segments?

Having two separate instructions for this seems reasonable and I like the new names, but could we choose a different opcode for `array.new_elem`? The problem with the current one is that it overlaps with V8's `array.init_from_data_static`, so implementing the new instruction with the opcode here will require a binary format back compat break in V8.

manoskouk: The current `array.new_from_data` in v8 supports only numeric arrays and data segments. Regarding `array.new_from_elem`, the discussion on #260 did not come up with a decision regarding how to handle cyclic dependencies between the element section and constant uses of `array.new_from_elem`, so we postponed its implementation. We could, of course, introduce it as a non-constant-only expression for now.

rossberg: @tlively, ah, sorry, I had missed the difference in operands. Changed opcodes to 0x1b and 0x1c, which seem to be free right now.

@manoskouk, thanks for the reminder. So far, the proposal has not declared any of the allocation instructions constant, although we discussed it. I suppose there is agreement to change this, so I went ahead and incorporated that into this PR. And I agree that, for the time being, excluding `array.new_elem` from being constant is the simplest interim solution. Once we have introduced repeated sections, we can relax that.

proposals/gc/Post-MVP.md
```diff
@@ -36,17 +36,50 @@ To that end, _bulk copying_ instructions could be added, similar to the [bulk in
   - `struct.copy $d : [(ref $d) (ref $s)] -> []` where both `$d` and `$s` are struct types, `$d` has only mutable fields, and `$s <: $d` modulo mutability
 
 * An instruction for bulk copying an array range:
-  - `array.copy $d : [(ref $d) i32 (ref $s) i32 i32] -> []`
-    - iff both `$d` and `$s` are array types
-    - and `$d` has mutable element type
-    - and `$s <: $d` modulo mutability
-  - the remaining operands are destination and source offset and length of the range
+  - `array.copy $d $s : [(ref null $d) i32 (ref null $s) i32 i32] -> []`
```
manoskouk, line 39: I think we should move `array.copy` to the MVP. J2CL has found it very useful in improving performance.

rossberg: I wouldn't be opposed to that, but it's a bit harder to argue that it is "MVP" material, given that it can be compiled out easily. Therefore, I would keep that decision separate from this PR, which focusses on an actual new ability, namely ways to create (useful) immutable arrays.

(Also, I'm not very keen on writing the test cases for such an instruction, so somebody else would have to own it as a PR. ;) )

proposals/gc/MVP.md
askeksa-google: FWIW, having these instructions as constant is essential for the dart2wasm use case of using them for constants. Including `array.new_elem` is necessary to not have a limit on how big `List` constants can be.

askeksa-google: `array.new_fixed` works fine as long as the created array does not exceed the maximum number of operands for `array.new_fixed`. For longer arrays, `array.new_elem` is currently the only way to create them as constant expressions.

tlively (approved):
Separately moving `array.copy` to the MPV sounds fine to me.

Issue #274, opened 2022-02-09
Refactor cast opcodes?


rossberg:
With the addition arrayref and structref, it would make sense to refactor cast and classification opcodes to something more modular: instead of a long list of individual opcodes, use an immediate, like in `br_on <heaptype>` (with some suitable restrictions), and similarly `ref.is <ht>` and `ref.as <ht>`. The interpreter has long been using something similar internally.

Edit: To clarify, this is solely suggesting a change to the encoding of this group of instructions, not affecting their semantics.

askeksa-google:
I was actually going to suggest exactly this, as part of a heap type refactoring based on an idea that came up in yesterday's discussion of [`i31` as a type dimension](https://github.com/WebAssembly/gc/issues/130) - specifically to have unions of heap types also be heap types. I ruminated about this idea and took it a bit further, and I think it could lead to a nice regularization of the type hierarchy - and, as you suggest, improved orthogonality between the type hierarchy and the instruction set.

I'll write my thoughts here in this issue, then, instead of opening a new one.

### The setup

The set of *singular heap types* includes a *hierarchy* of abstract (`any`, `func`, `eq`, `data`, `array`, `struct`) and concrete (defined functions, arrays and structs) pointer types, plus two *isolated* heap types - `null` and `i31`.

A heap type is a union of zero or more singular heap types. As a start, we can restrict such unions to include at most one hierarchy type. We can the generalize them later if needed.

We will have a syntax for expressing a union. There can be constructors for the combinations of `null` and `i31`, followed by a hierarchy type. When a type (hierarchy or isolated) is mentioned without a constructor, we'll need some conventions for which of the other components are included by default (like `null` is with the current shorthands).

### Testing instructions

With such a setup, we can replace our plethora of testing instructions (`ref.is_*`, `ref.as_*`, `br_on_*`, at least 24 instructions in total, currently, counting only static versions) by just 5: `ref.is`, `ref.as`, `ref.as_not`, `br_on` and `br_on_not`.

[...]

PR #284, opened 2022-03-17
Remove the immediate from array.len in binary format

titzer:
This addresses a small TODO in the binary encoding/decoding of the reference interpreter.


tlively:
lgtm fwiw

rossberg (approved):
LGTM, but FWIW, I left this in intentionally to remain compatible with V8 until they decide to update their implementation. Has that happened? Otherwise, e.g. Wob&Waml will break.

manoskouk:
We have not updated the V8 implementation yet. We are planning to implement breaking changes when other discussions are also resolved (e.g. #274).

titzer:
It's already a breaking change, because the reference interpreter is using a zero byte now, and V8 would be interpreting that as a type index.

@manoskouk This should be a very small number of lines to change in V8. Any chance we can go ahead with it, without blocking on another discussion?

manoskouk:
In V8, we [accept but ignore](https://docs.google.com/document/d/1DklC3qVuOdLHSXB5UXghM_syCh-4cMinQ50ICiXnK3Q/edit#bookmark=id.m372kw3q8qfn) the type index.
This is compatible with the code containing a zero byte instead. If we remove it completely, it will break our partners' compiled benchmarks.
That said, we understand that the online spec's changes cannot always be synced to ours; this is why we maintain a separate spec for V8's implementation. So feel free to merge this change, just please be aware that our implementation will adopt it later.

titzer:
Ok, thanks for the explanation.

We'll have to start breaking binary compatibility, either little by little or all at once. As this is a small one, I think we should land it, but I'll let @rossberg make the call, since he'll have to update Wob and Waml.

rossberg:
The reference interpreter currently also decodes but ignores the type index.

The problem with Wob and Waml isn't that I have to update them, but that their test suite runs on both the reference interpreter and V8. So if they diverge, breakage is unavoidable.

Do you have a pressing need to land this now?

PR #286, opened 2022-03-24
Additional negative subtyping tests

titzer:
This PR adds additional tests for declarations that attempt to define structs, arrays, and function types that have invalid subtyping relations.

### DIFF
--- a/test/core/gc/type-subtyping.wast
+++ b/test/core/gc/type-subtyping.wast
@@ -134,3 +134,79 @@
   )
 )
 (assert_return (invoke "run"))
+
+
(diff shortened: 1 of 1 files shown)


tlively (approved):

tlively:
Does "sub type" already exist as a reason for validation failure? If it is not already established, perhaps "subtype" would be better.

titzer:
"sub type" matches the beginning of the error string generated by the reference interpreter for these test cases. I agree that "subtype" would be better, but we can do that in a separate change.

rossberg (approved):

PR #295, opened 2022-05-04
Add nullref

rossberg:
Bring back the nullref type.

- In binary, it's a heaptype encoded as -0x12 (previously used for anyref, now unused).
- In text, `nullref` is shorthand for `(ref null none)`, where `none` is the name for the bottom heap type (chosen as opposite to `any`).

No changes needed to the interpreter other than supporting it in binary and text format and allowing it in the validator. The rest of the change is merely renaming and reordering.

### DIFF
--- a/interpreter/binary/decode.ml
+++ b/interpreter/binary/decode.ml
@@ -173,12 +173,12 @@ let heap_type s =
     (match vs7 s with
     | -0x10 -> FuncHeapType
     | -0x11 -> AnyHeapType
-    | -0x12 -> NoneHeapType
     | -0x13 -> EqHeapType
     | -0x16 -> I31HeapType
     | -0x18 -> RttHeapType (var_type s)
     | -0x19 -> DataHeapType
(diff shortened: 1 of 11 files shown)


titzer:
I think it'd be best to avoid reusing the code for `anyref` just yet., in particular we may want to revisit https://github.com/WebAssembly/gc/pull/271. That introduced a mode where an engine still has to model `externref` to pass the old spec and also model `anyref` to pass the GC spec.

rossberg:
@titzer, if you need that mode flag anyway, wouldn't it trivially enable interpreting the code accordingly as well?

titzer:
@rossberg I think it will be safer to use a new code to separate this PR (which is completely reasonable) from that other discussion.

rossberg:
Okay, I changed the code to 0x1b for now, assuming that we'll eventually do a code space cleanup anyway. PTAL.

titzer (approved):
lgtm

PR #301, opened 2022-05-25
Add example for RTTs + generics

tlively:
Concretize the argument that implicit RTTs would have unnecessary overhead by providing an example use case, generic containers, that would not need RTTs.


rossberg (approved):
Thanks!

PR #306, opened 2022-06-08
Move RTTs to Post-MVP

rossberg:
This PR removes explicit RTTs and moves them to Post-MVP. Updates design docs, interpreter, and tests.

The `rtt $t` type and the `rtt.canon` instruction are removed and all RTT-consuming instructions (allocations and casts) are replaced with "canonical" versions that embed the semantics of `rtt.canon`. RTT-dependent cast instructions now need a static type as an additional immediate.

Obviously, this is a breaking change across the board, so I did not attempt to maintain backwards compat in details such as opcode assignments this time.


interpreter/README.md
```diff
@@ -254,13 +254,13 @@ op:
   br_table <var>+
   br_on_null <var>
   br_on_<castop> <var>
-  br_on_cast <var>
+  br_on_cast_canon <var> (type <var>)
   br_on_non_null <var>
   br_on_non_<castop> <var>
-  br_on_cast_fail <var>
+  br_on_cast_canon_fail <var> (type <var>)
```
tlively, line 260: What is `(type <var>)` here? Why isn't it added to other instructions like `struct.new` below?

proposals/gc/Post-MVP.md
```diff
@@ -366,6 +367,110 @@ Interior references:
...
+* Runtime types are explicit values representing concrete types at runtime; a value of type `rtt <typeidx>` is a dynamic representative of the static type `<typeidx>`.
+
+* All RTTs are explicitly created and all operations involving dynamic type information (like casts) operate on explicit RTT operands.
+
+This will require adding a new form of heap type:
+
+* `rtt <typeidx>` is a new heap type that is a runtime representation of the static type `<typeidx>`
+  - `heaptype ::= ... | rtt <typeidx>`
+  - `rtt t ok` iff `t ok`
+  - the reference type `(rtt $t)` is a shorthand for `(ref (rtt $t))`
```
tlively, line 383: Would this be a text-only shorthand? It seems slightly inconsistent that this shorthand corresponds to a non-nullable reference while our other shorthands correspond to nullable references, but I don't think that's a big deal.

tlively (commented):
Thanks for doing this!

My one big comment is that I think it would be better for the MVP instructions not to contain `_canon` in their names. Understanding what `_canon` means would require knowing some of the context of the development of the proposal, but ideally the proposal would not require such context to fully understand. Also, having `_canon` in the name seems to imply that the instructions will always work for any implicit `rtt.canon`, but I think it would make sense to leave ourselves space to consider whether the sugar should be restricted to only a subset of possible `rtt.canon` cases. I would prefer to differentiate the post-MVP instructions by adding `_rtt` to their names instead.

interpreter/README.md
rossberg: This is the type to cast to, as it can no longer be derived from the RTT operand. Most other instructions already had a (previously redundant) explicit type (cf the discussion regarding type annotations).

The syntax is chosen to mirror other cases like `call_indirect`, and to avoid confusion with the branch label.

I'd be fine with switching to `(type _)` syntax everywhere, it would be more consistent – but also more verbose. I justified the difference to myself by the prevalent choice that indices associated with the primary object kind of an instruction do not have a sort prefix (e.g., the global index in `global.get`, the table index in `call_indirect`, the label index in branches) whereas secondary ones do (e.g., the type index in `call_indirect` and `br_on_cast`, the memory index in `data`). And one can argue that the type in GC instructions represents the primary object kind, if that makes sense. :)

interpreter/syntax/ast.ml
```diff
@@ -127,7 +127,7 @@ type vec_storeop = (vec_type, unit) memop
 type vec_laneop = (vec_type, pack_size) memop * int
 
 type initop = Explicit | Implicit
-type castop = NullOp | I31Op | DataOp | ArrayOp | FuncOp | RttOp
+type castop = NullOp | I31Op | DataOp | ArrayOp | FuncOp | RttOp of int32 Source.phrase
```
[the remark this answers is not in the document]
rossberg: See above, this is the now necessary type immediate.

proposals/gc/Post-MVP.md
rossberg: It's what we previously had, including in the binary format. I agree it's a minor inconsistency, but you rarely ever want nullable RTTs.

proposals/gc/Post-MVP.md
```diff
@@ -366,6 +367,110 @@ Interior references:
...
+At a baseline, RTT values can be created with a new instruction:
+```
+* `rtt.canon <typeidx>` returns the RTT of the specified type
+  - `rtt.canon $t : [] -> [(rtt $t)]`
+  - multiple invocations of this instruction yield the same observable RTTs
+  - this is a *constant instruction*
+```
+With extensions like [type parameters](#type-parameters), this instruction will become more nuanced and will involve additional RTT operands if `$t` has type paramters.
+
+Correspondingly, all allocation and cast instructions will get counter parts taking an explicit RTT operand. The MVP's canonical versions can be reinterpreted as the combination of the explicit ones whose RTT operand is created with `rtt.canon`:
```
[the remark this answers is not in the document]
rossberg: I hope not, as that would create a lot of extra complexity in the instruction set and also be somewhat inconsistent (the primary RTT operand is implicit but the others are not?).

But it is this kind of potential extra complexity and non-trivial warts that makes me so uneasy. ;)

rossberg (commented):
The idea is that the generic mnemonics ought to be reserved for the instructions that can actually handle the general case. I’m not wed to the particular choice of suffix (better suggestions welcome!), but I do think that the mnemonics need to reflect that these instructions are limited to a special case in the future.

jakobkummerow:
Regarding opcodes, since we've had the RTT-free instructions implemented in both V8 and Binaryen for a long time now, we also have an existing set of opcode assignments (that don't overlap with the other opcodes), see [here](https://docs.google.com/document/d/1OcajmqehEBwmghdQJOtcOjaeBh4QHTlQFRseEVl0hUM/edit?pli=1#heading=h.9hpf35cp6uqr). It would be nice to maintain them for now? Not a big issue though, we'll need to make some backwards-incompatible changes soon anyway and could batch up the opcode changes with those.

The state in that doc is in part due to this proposal's history: you may recall that in days long gone we actually had both `struct.new` and `struct.new_with_rtt`, then at some point dropped the former, then (in V8/Binaryen at least) reintroduced it, now we're dropping the latter. This history also explains the textual names used there: for the allocating opcodes, `.new` is with implicit RTT and `.new_with_rtt` is with explicit RTT. We've never renamed them to preserve backwards compatibility of the text format. For the cast/test instructions, there is no such history, so we picked the `_static` (as in: "statically typed") suffix when we added the implicit-RTT versions. (I don't care too much about renaming some or all of the instructions, and it would certainly be nice to make them consistent; I'm just explaining how we arrived at what that doc is saying today.)

---

A possible (follow-up?) simplification could be to merge `ref.cast_canon`, `ref.test_canon`, `br_on_cast_canon` and `br_on_cast_canon_fail` into `ref.as <ht>`, `ref.is <ht>`, `br_on <ht>` and `br_on_not <ht>`, respectively, as issue #274 sketches out. I don't see a reason to keep them separate, and merging them might help assuage concerns people have about the overall size of the instruction set once we re-introduce explicit-RTT versions later. That said, I don't feel strongly about it, and if there are reasons to keep them separate, we can certainly do that too.
(The fact that the `*_canon` group currently takes a `dataref` rather than an `anyref` as input is not providing any benefits, as discussed at length on that issue, so that's not a reason to keep the distinction. Effectively, to cast to type $T, you need to use `ref.as` iff $T is generic and `ref.cast_canon` iff $T is non-generic, and I don't see this distinction providing any practical benefits. There is no $T for which `ref.as $T` and `ref.cast_canon $T` are both valid, so no expressiveness is lost if we merge them.)

rossberg:
Yes, refactoring the cast operators is a follow-up. But it's gonna be affected by the anyref/funcref/externref subtyping questions, so it makes more sense to wait with it until we have resolves that.

rossberg:
To avoid conflicts with followup PRs, taking the liberty to land this. We can bikeshed the mnemonics later.

PR #310, opened 2022-07-14
3-pronged subtyping hierarchy

rossberg:
This makes the necessary changes to the MVP doc for introducing the 3-pronged subtype hierarchy. For now, I haven’t done any renaming, so I went with types `any` and `data` and leave possible bikeshedding for later. Concretely:

* Reintroduce heap type `any` separately from `extern` (under the previous type code)
* Add two new bottom types `nofunc` and `noextern`
* Add shorthands `nullfuncref = (ref null nofunc)` and `nullexternref = (ref null noextern)`
* Split up subtyping rules for top and bottom types; they now need to inspect `t`, unlike before
  (in other words, checking against `any` becomes more costly)
* Extend typing rules for casts to check that they are hierarchy-consistent
* Add instructions `extern.internalize` and `extern.externalize`
* Remove instructions `ref.is_func`, `ref.as_func`, `br_on_func`, `br_on_non_func`

As discussed, there are no instructions for converting between extern and functions.

See #311 for implementation and tests.


rossberg:
See #311 for implementation and tests.

manoskouk:
I still believe we do not need `noextern`. We can include `(ref.null extern)` in `externref`.

proposals/gc/MVP.md
```diff
@@ -529,10 +567,22 @@ Tentatively, support a type of guaranteed unboxed scalars.
   - traps if the operand is null
 
 
-#### Classification
+#### External conversion
+
+* `extern.internalize` converts an external value into the internal representation
```
manoskouk, line 572: In the discussions, in was mentioned that it would be useful for performance to include direct transformations to more specific abstract types. This might be best as a post-MVP feature though.

manoskouk:
A while back @rossberg suggested introducing a `struct` type for consistency with `array`. It was argued then (including by myself) that this does not add anything to the language.
However, it might be useful to include it because downcasts from `struct` are likely to be more efficient than from `data`, and even more so the sequence `ref.as_struct (ref.cast $t)` compared to `ref.as_data (ref.cast $t)`.
Again, this might be a post-MVP consideration.

rossberg:
I'm all for adding `struct`, even in the MVP, to avoid an odd asymmetry. But I didn't want to mix it up with this PR.

manoskouk:
> @manoskouk, I believe that wouldn't be forward-compatible. For example, if we later introduce type imports `$t <: extern` like @lukewagner mentioned, then `(ref null extern)` would no longer be a correct type for the null reference. Consequently, we would need to change `nullexternref` to mean something else later. But that would break existing programs that assume `(ref null extern)` is the same type (or a subtype).

I am not following. Aren't we doing the same thing for `func` right now? `(ref.null func)` is typed `(ref null func)` in the MVP. Now that we introduced subtypes of `func`, we are retyping it as `(ref null nofunc)`. Why can we not follow the same process for `externref` when and if we introduce subtypes of it?

rossberg:
> Aren't we doing the same thing for `func` right now? `(ref.null func)` is typed `(ref null func)` in the MVP.

Only if you are assuming that we are removing `nullexternref` as well. Then we'd indeed be in a similar situation as with func today. But we're only in that because we had no nullref types in the first place. We have identified that as a mistake and are rectifying it now, so it seems preferable to do so consistently, and have explicit null types for every hierarchy. Otherwise we'd still provide no forward-compatible way to type-annotate an external nullref. Stupid example:
```
(module
  (global (export "null") externref (ref.null extern))
)
```
This export will no longer be adequate in the presence of additional subtypes of extern. Maybe such a case is not very relevant in practice, but I see no benefit in not rectifying it along with the others. It's not like omitting it would simplify much.

proposals/gc/MVP.md
```diff
@@ -688,14 +719,17 @@ This extends the [encodings](https://github.com/WebAssembly/function-references/
...
+| -0x11  | `externref`     |            | shorthand, from reftype proposal |
+| -0x12  | `anyref`        |            | shorthand |
 | -0x13  | `eqref`         |            | shorthand |
 | -0x14  | `(ref null ht)` | `ht : heaptype (s33)` | from funcref proposal |
 | -0x15  | `(ref ht)`      | `ht : heaptype (s33)` | from funcref proposal |
 | -0x16  | `i31ref`        |            | shorthand |
 | -0x19  | `dataref`       |            | shorthand |
 | -0x1a  | `arrayref`      |            | shorthand |
 | -0x1b  | `nullref`       |            | shorthand |
+| -0x1c  | `nullfuncref`   |            | shorthand |
```
jakobkummerow, line 731: Note that `-0x1c` through `-0x1f` are taken by the stringrefs proposal, see https://github.com/WebAssembly/stringref/blob/main/proposals/stringref/Overview.md#binary-encoding.

rossberg: Thanks for pointing out. I opened WebAssembly/stringref#41.

jakobkummerow: FYI, we've implemented these as:
```
| -0x17 | `nullexternref` |
| -0x18 | `nullfuncref`   |
```
to resolve the conflict with the stringref proposal. (Of course this applies to lines 749/750 below as well.)

jakobkummerow (approved):
Generally LGTM.

I'd recommend to make the binary encoding listed here reflect reality (see note below) before landing this PR, but that's up to you.

rossberg:
Changed opcodes.

In the future, I would appreciate if we could get away from this mode of operation where implementations unilaterally deviate from what's proposed and then retroactively declare that "reality". ;)

tlively:
> In the future, I would appreciate if we could get away from this mode of operation where implementations unilaterally deviate from what's proposed and then retroactively declare that "reality". ;)

What other workflow would you want to see that still allows engines to have multiple arbitrary prototype/early-stage proposals in flight at once?

rossberg:
I think it's less a question of workflow than one of style and communication. When you discover conflicts between, or unresolved issues with, proposals, try to help the champions to get them resolved. If resolution is blocking you for too long, discuss what temporary choices you could move forward with. Consider them temporary.

PR #322, opened 2022-08-17
Remove tests for func.bind

titzer:
@tlively Trying to bring the state of tests in this repo up to date.

This PR simply removes a test for `func.bind`, which has been removed from the `function-references` proposal. This test has already been removed from that repo. Technically the proper way is to merge this proposal with changes from function-references, but that is a more involved process that is likely blocked on bringing function-references up to date.

As such, this is an easy change and should be easy to merge when function-references is merged back in which is probably best done by @rossberg .


tlively (commented):
SGTM 👍

titzer:
sgtm is not an official approval, but this is a fairly simple change. Should I merge this?

tlively (approved):

tlively:
Yes, I think so.

rossberg:
For future reference, the preferred way to sync with changes from an upstream repo is to re-merge the upstream repo. ;)

PR #325, opened 2022-09-21
Refactor casts

rossberg:
Addresses #274.

This PR modifies the MVP doc to replace the existing classification and cast instructions with casts that can target any reference type:
```
ref.test null? <heaptype>
ref.cast null? <heaptype>
br_on_cast $l null? <heaptype>
br_on_cast_fail $l null? <heaptype>
```
The typing rules for these casts become somewhat complex, in order (1) to make sure that source and target type are in the same hierarchy, and (2) to handle the propagation and negation of the optional `null` attributes correctly.

Furthermore, the PR repurposes the `dataref` type as `structref`. The main purpose of `data` was to denote the required source type for concrete casts. Since that requirement is gone, so is the purpose. For the remaining use cases, `struct` is typically more precise. That also avoids some of the concerns that came up earlier regarding the forward compatibility of exhaustive type inspections on such a "middle" type.

Note: I gave up on the idea of having two separate cast instructions for abstract vs concrete types, since that just seemed a bit silly at this point – I convinced myself that the distinction can be reintroduced later in a warty way if needs be. I don't expect that anybody will object to that. I also included top types and `eqref` as allowable cast targets, although I still think this actually is a directional mistake and introduces both a legacy cost and creates the wrong expectations moving forward. Mainly I've just grown tired of the discussion.

### DIFF
--- a/proposals/gc/MVP.md
+++ b/proposals/gc/MVP.md
@@ -600,12 +600,12 @@ Casts work for both abstract and concrete types. In the latter case, they test i
 
 * `ref.test null? <heaptype>` checks whether a reference has a given heap type
   - `ref.test null? ht : [(ref null ht')] -> [i32]`
-    - iff `ht <: ht'`
+    - iff `ht <: tht` and `ht' <: tht` where `tht` is a common super type
   - if `null?` is present, returns 1 for null, otherwise 0
 
 * `ref.cast null? <heaptype>` tries to convert to a given heap type
   - `ref.cast null? ht : [(ref null ht')] -> [(ref null2? ht)]`
-    - iff `ht <: ht'`
+    - iff `ht <: tht` and `ht' <: tht` where `tht` is a common super type
     - and `null? = null2?`
   - traps if reference is not of requested type
   - if `null?` is present, a null operand is passed through, otherwise traps on null
@@ -629,7 +629,7 @@ Casts work for both abstract and concrete types. In the latter case, they test i
   - passes operand along with branch, plus possible extra args
   - if `null?` is present, does not branch on null, otherwise does
 
-Note: Via subsumption to a common supertype, `ref.test` and `ref.cast` actually allow an operand of sibling reference type, even though this cannot possibly succeed. In the case of the branch instruction, this has to be made explicit, though, since the type also occurs in result position. The given typing rules allow this type to also propagate to the result, which avoids the need to compute a least upper bound between the operand type and the target type in the typing algorithm (allowing so is required to maintain subtype substitutability). In practice, it is sufficient to test that both types share the same top heap type.
+Note: Cast instructions do _not_ require the operand's source type to be a supertype of the target type. It can also be a "sibling" in the same hierarchy, i.e., they only need to have a common supertype (in practice, it is sufficient to test that both types share the same top heap type.). Allowing so is necessary to maintain subtype substitutability, i.e., the ability to maintain well-typedness when operands are replaced by subtypes.
 
 Note: The [reference types](https://github.com/WebAssembly/reference-types) and [typed function references](https://github.com/WebAssembly/function-references)already introduce similar `ref.is_null`, `br_on_null`, and `br_on_non_null` instructions. These can now be interpreted as syntactic sugar:
 
--- a/proposals/gc/MVP.md
+++ b/proposals/gc/MVP.md
@@ -598,48 +598,48 @@ In particular, `ref.null` is typed as before, despite the introduction of `none`
 
 Casts work for both abstract and concrete types. In the latter case, they test of the operands RTT is a sub-RTT of the target type.
 
-* `ref.is null? <heaptype>` checks whether a reference is a given heap type
-  - `ref.is null? ht : [(ref null tht)] -> [i32]`
-    - iff `ht <: tht` and `tht` is a top heap type
+* `ref.test null? <heaptype>` checks whether a reference is a given heap type
+  - `ref.test null? ht : [(ref null ht')] -> [i32]`
+    - iff `ht <: tht` and `ht' <: tht` where `tht` is a common super type
   - if `null?` is present, returns 1 for null, otherwise 0
 
-* `ref.as null? <heaptype>` tries to convert to a given heap type
-  - `ref.as null? ht : [(ref null tht)] -> [(ref null2? ht)]`
-    - iff `ht <: tht` and `tht` is a top heap type
+* `ref.cast null? <heaptype>` tries to convert to a given heap type
+  - `ref.cast null? ht : [(ref null ht')] -> [(ref null2? ht)]`
+    - iff `ht <: tht` and `ht' <: tht` where `tht` is a common super type
     - and `null? = null2?`
   - traps if reference is not of requested type
   - if `null?` is present, a null operand is passed through, otherwise traps on null
-  - equivalent to `(block $l (param anyref) (result (ref null? ht)) (br_on null? ht $l) (unreachable))`
+  - equivalent to `(block $l (param anyref) (result (ref null? ht)) (br_on_cast null? ht $l) (unreachable))`
 
-* `br_on <labelidx> null? <heaptype>` branches if a reference is given heap type
-  - `br_on $l null? ht : [t0* (ref null ht')] -> [t0* (ref null2? ht')]`
+* `br_on_cast <labelidx> null? <heaptype>` branches if a reference is given heap type
+  - `br_on_cast $l null? ht : [t0* (ref null ht')] -> [t0* (ref null2? ht')]`
     - iff `$l : [t0* t']`
     - and `(ref null3? ht) <: t'`
-    - and `ht <: tht` and `ht' <: tht` and `tht` is a top heap type
+    - and `ht <: tht` and `ht' <: tht` where `tht` is a common super type
     - and `null? = null3? =/= null2?`
   - passes operand along with branch under target type, plus possible extra args
   - if `null?` is present, branches on null, otherwise does not
 
-* `br_on_non <labelidx> null? <heaptype>` branches if a reference is not a given heap type
-  - `br_on_non $l null? ht : [t0* (ref null ht')] -> [t0* (ref null2? ht)]`
+* `br_on_cast_fail <labelidx> null? <heaptype>` branches if a reference is not a given heap type
+  - `br_on_cast_fail $l null? ht : [t0* (ref null ht')] -> [t0* (ref null2? ht)]`
     - iff `$l : [t0* t']`
     - and `(ref null3? ht) <: t'`
-    - and `ht <: tht` and `ht' <: tht` and `tht` is a top heap type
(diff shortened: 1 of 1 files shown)


proposals/gc/MVP.md
```diff
@@ -594,112 +594,52 @@ In particular, `ref.null` is typed as before, despite the introduction of `none`
...
-  - passes operand along with branch as data, plus possible extra args
+#### Casts
 
-* `br_on_non_data <labelidx>` branches if a reference is not compound data
-  - `br_on_non_data $l : [t0* t] -> [t0* (ref data)]`
-    - iff `$l : [t0* t']`
-    - and `t <: anyref`
-    - and `t <: t'`
-  - passes operand along with branch, plus possible extra args
+Casts work for both abstract and concrete types. In the latter case, they test of the operands RTT is a sub-RTT of the target type.
```
jakobkummerow, line 599:
```suggestion
Casts work for both abstract and concrete types. In the latter case, they test if the operand's RTT is a sub-RTT of the target type.
```

proposals/gc/MVP.md
```diff
@@ -594,112 +594,52 @@ In particular, `ref.null` is typed as before, despite the introduction of `none`
...
-  - `ref.as_array : [anyref] -> [(ref array)]`
-  - traps if reference is not an array
-  - equivalent to `(block $l (param anyref) (result (ref array)) (br_on_array $l) (unreachable))`
+Note: The [reference types](https://github.com/WebAssembly/reference-types) and [typed function references](https://github.com/WebAssembly/function-references)already introduce similar `ref.is_null`, `br_on_null`, and `br_on_non_null` instructions. These can now be interpreted as syntactic sugar:
 
-* `ref.as_i31` converts to an integer reference
-  - `ref.as_i31 : [anyref] -> [(ref i31)]`
-  - traps if reference is not an integer
-  - equivalent to `(block $l (param anyref) (result (ref i31)) (br_on_i31 $l) (unreachable))`
+* `ref.is_null` is equivalent to `ref.test null ht`, where `tht` is the suitable bottom type (`none`, `nofunc`, or `noextern`)
```
jakobkummerow, line 636: `tht` doesn't show up before "where". Did you mean `ht`?

Same below (lines 638, 640).

proposals/gc/MVP.md
```diff
@@ -594,112 +594,52 @@ In particular, `ref.null` is typed as before, despite the introduction of `none`
...
+  - `ref.cast null? ht : [(ref null ht')] -> [(ref null2? ht)]`
+    - iff `ht <: tht` and `ht' <: tht` where `tht` is a common super type
+    - and `null? = null2?`
+  - traps if reference is not of requested type
+  - if `null?` is present, a null operand is passed through, otherwise traps on null
+  - equivalent to `(block $l (param anyref) (result (ref null? ht)) (br_on_cast null? ht $l) (unreachable))`
 
-* `br_on_i31 <labelidx>` branches if a reference is an integer
-  - `br_on_i31 $l : [t0* t] -> [t0* t]`
+* `br_on_cast <labelidx> null? <heaptype>` branches if a reference is given heap type
```
jakobkummerow, line 614:
```suggestion
* `br_on_cast <labelidx> null? <heaptype>` branches if a reference is a given heap type
```

proposals/gc/MVP.md
```diff
@@ -817,22 +757,14 @@ The opcode for heap types is encoded as an `s33`.
...
-| 0xfb62 | `br_on_i31` | `$l : labelidx` |
-| 0xfb64 | `br_on_non_data` | `$l : labelidx` |
-| 0xfb65 | `br_on_non_i31` | `$l : labelidx` |
-| 0xfb66 | `br_on_array` | `$l : labelidx` |
-| 0xfb67 | `br_on_non_array` | `$l : labelidx` |
+| 0xfb40 | `ref.test ht` | `ht : heaptype` |
+| 0xfb41 | `ref.cast ht` | `ht : heaptype` |
+| 0xfb42 | `br_on_cast $l ht` | `$l : labelidx`, `ht : heaptype` |
+| 0xfb43 | `br_on_cast_fail $l ht` | `$l : labelidx`, `ht : heaptype` |
+| 0xfb40 | `ref.test null ht` | `ht : heaptype` |
```
jakobkummerow, line 764: Did you mean `0xfb50..0xfb53` for this and the next three? Another good option would be the 48..4b range for the `null`-enabled set.

(You may be tempted to use the range 44..47, but that is [currently](https://docs.google.com/document/d/14xWenN0lUoHa9TiDJd8yYR8ZKjgOhgO7iEmfrAuA1_M/edit#) in use in implementations, and we have a requirement not to cause breakage in the ongoing test coverage of our prototyping partners, so repurposing those would cause some logistical overhead because we'd have to do a back-and-forth migration dance, so I'd prefer to use a range that's currently not in use. Since we'll sort/compact/cleanup the encoding space before finalizing the proposal anyway, it's acceptable to have cases of minor ugliness for the time being, such as non-contiguous ranges for similar instructions.
For reference, you can always look up what prototype implementations across the stack (J2Wasm, Kotlin-Wasm, Dart2Wasm, Binaryen, V8, and I believe Spidermonkey too) are doing at https://bit.ly/3cWcm6Q, we've been keeping that link stable for the past 2.5 years and will continue to keep it. Links to drafts for updated versions, like the one linked above, are always at the top of that main document.)

jakobkummerow (commented):
I like it. Only spotted a few typos.

proposals/gc/MVP.md
```diff
@@ -594,112 +594,52 @@ In particular, `ref.null` is typed as before, despite the introduction of `none`
...
 
-* `br_on_non_array <labelidx>` branches if a reference is not an array
-  - `br_on_non_array $l : [t0* t] -> [t0* (ref array)]`
-    - iff `$l : [t0* t']`
-    - and `t <: anyref`
-    - and `t <: t'`
-  - passes operand along with branch, plus possible extra args
+* `ref.cast null? <heaptype>` tries to convert to a given heap type
+  - `ref.cast null? ht : [(ref null ht')] -> [(ref null2? ht)]`
+    - iff `ht <: tht` and `ht' <: tht` where `tht` is a common super type
```
conrad-watt, line 608: Just for my own understanding, is `ht <: ht'` an equivalent formulation of this condition, or is there some subtlety around subsumption that I don't have right in my head?

proposals/gc/MVP.md
```diff
@@ -594,112 +594,52 @@ In particular, `ref.null` is typed as before, despite the introduction of `none`
...
+    - and `null? = null2?`
+  - traps if reference is not of requested type
+  - if `null?` is present, a null operand is passed through, otherwise traps on null
+  - equivalent to `(block $l (param anyref) (result (ref null? ht)) (br_on_cast null? ht $l) (unreachable))`
 
-* `br_on_i31 <labelidx>` branches if a reference is an integer
-  - `br_on_i31 $l : [t0* t] -> [t0* t]`
+* `br_on_cast <labelidx> null? <heaptype>` branches if a reference is given heap type
+  - `br_on_cast $l null? ht : [t0* (ref null ht')] -> [t0* (ref null2? ht')]`
```
conrad-watt, line 615: Should `ref null ht'` in the input type be `ref null2? ht'`?

proposals/gc/MVP.md
```diff
@@ -594,112 +594,52 @@ In particular, `ref.null` is typed as before, despite the introduction of `none`
...
-  - equivalent to `(block $l (param anyref) (result (ref i31)) (br_on_i31 $l) (unreachable))`
+* `ref.is_null` is equivalent to `ref.test null ht`, where `tht` is the suitable bottom type (`none`, `nofunc`, or `noextern`)
 
-Note: The [reference types](https://github.com/WebAssembly/reference-types) and [typed function references](https://github.com/WebAssembly/function-references)already introduce similar `ref.is_null`, `br_on_null`, and `br_on_non_null` instructions.
-
-Note: The `br_on_*` instructions allow an operand of unrelated reference type, even though this cannot possibly succeed. That's because subtyping allows to forget that information, so by the subtype substitutibility property, it would be accepted in any case. The given typing rules merely allow this type to also propagate to the result, which avoids the need to compute a least upper bound between the operand type and the target type in the typing algorithm.
-
-
-#### Casts
+* `br_on_null` is equivalent to `br_on_cast null ht`, where `tht` is the suitable bottom type
```
conrad-watt: Should "is the suitable bottom type" here instead be "is the heap type of the operand" (c.f. `ref.as_non_null` below).

proposals/gc/MVP.md
rossberg: No, the input type can always allow null, analogous to other ref instructions. You may indeed want a non-null cast on an operand that you know can't actually be null, but you can't prove it to the Wasm type system.

proposals/gc/MVP.md
rossberg: No, br_on_null effectively is a downcast to null, with the other side of the union being empty.

But I just realised that br_on_(non)_null do not pass on the null value, so they are not exactly equivalent. Tweaked text.

proposals/gc/MVP.md
conrad-watt: makes sense, thanks!

rossberg:
An implementation in the reference interpreter and respective tests can be found in #326.

proposals/gc/MVP.md
```diff
@@ -594,112 +594,54 @@ In particular, `ref.null` is typed as before, despite the introduction of `none`
...
   - passes operand along with branch, plus possible extra args
+  - if `null?` is present, does not branch on null, otherwise does
 
-* `br_on_array <labelidx>` branches if a reference is an array
-  - `br_on_array $l : [t0* t] -> [t0* t]`
-    - iff `$l : [t0* t']`
-    - and `t <: anyref`
-    - and `(ref array) <: t'`
-  - passes operand along with branch as data, plus possible extra args
+Note: Via subsumption to a common supertype, `ref.test` and `ref.cast` actually allow an operand of sibling reference type, even though this cannot possibly succeed. In the case of the branch instruction, this has to be made explicit, though, since the type also occurs in result position. The given typing rules allow this type to also propagate to the result, which avoids the need to compute a least upper bound between the operand type and the target type in the typing algorithm (allowing so is required to maintain subtype substitutability). In practice, it is sufficient to test that both types share the same top heap type.
```
tlively, line 632: I don't understand the second or third sentences of this note. What has to be made explicit and what type are they referring to?

tlively (approved):
LGTM besides that one request for clarification.

proposals/gc/MVP.md
conrad-watt: @rossberg to be honest, it might be better to go back to the previous formulation of the side condition, even though `ref.test` and `ref.cast` can have a simpler formulation than `br_on_*`, just to avoid needing to explain that the conditions for all three instructions are "really" the same.

rossberg: Okay, I reverted the condition and simplified the note accordingly. Dropped the bit about the lub, since it is somewhat tangential and probably more confusing than helpful.

PR #330, opened 2022-10-21
Allow mutable fields in struct.new[_default]

kripken:
The mutability or immutability of fields should not matter when creating a struct, unless I am missing something? That is, it's always fine to create a struct (with or without default values) even if the fields are immutable. Immutability just affects changes later, using `struct.set`:

```wat
(type $T (struct_subtype (field i32) data)

(struct.new_default $T) ;; this is fine, the field will contain 0, and never be modified later.
(struct.set $T 0 ..) ;; this is not fine, the field is immutable.
```

This was noticed on a [Dart benchmark](https://github.com/dart-lang/sdk/issues/50268). V8 atm will error on immutable fields, following the old text here, and Binaryen will turn fields immutable when it sees no sets to them.

edit: I was confused about the `canon` names, sorry about that. PR has been updated.

### DIFF
--- a/proposals/gc/MVP.md
+++ b/proposals/gc/MVP.md
@@ -487,13 +487,13 @@ In particular, `ref.null` is typed as before, despite the introduction of `none`
 
 * `struct.new_canon <typeidx>` allocates a structure with canonical [RTT](#values) and initialises its fields with given values
   - `struct.new_canon $t : [t'*] -> [(ref $t)]`
-    - iff `expand($t) = struct (mut t'')*`
+    - iff `expand($t) = struct (mut? t'')*`
     - and `(t' = unpacked(t''))*`
   - this is a *constant instruction*
 
 * `struct.new_canon_default <typeidx>` allocates a structure of type `$t` with canonical [RTT](#values) and initialises its fields with default values
   - `struct.new_canon_default $t : [] -> [(ref $t)]`
-    - iff `expand($t) = struct (mut t')*`
+    - iff `expand($t) = struct (mut? t')*`
     - and all `t'*` are defaultable
   - this is a *constant instruction*
 


tlively (approved):
This makes sense to me.

rossberg:
Ah, you may be misparsing the rule: as in the spec, `mut` here is a placeholder variable for the mutability attribute, which may be instantiated with either `var` or `const` -- though that's admittedly hard to see without appropriate type-setting, but cf how it's used in other rules. So it already says what you think it should be saying. :)

kripken:
@rossberg I see, thanks! Sounds good then, closing.

tlively:
That’s confusing since var is written `mut` in the text format. Can we change something to make this clearer?

rossberg:
@tlively, I agree, and wouldn't mind adjusting the spec. Do you have a good suggestion what to substitute "mutability" with?

tlively:
I think "mutability" is the right name for the attribute. Could we rename "var" to "mutable" and possibly "const" to "immutable"?

dcodeIO:
Could perhaps spell in full for clarity, so `mutable` and `mutability` can be distinguished?

* iff `expand($t) = struct (mutability t')*`

Seems this is done in other places as well:

* `fieldtype ::= <mutability> <storagetype>`

rossberg:
> I think "mutability" is the right name for the attribute. Could we rename "var" to "mutable" and possibly "const" to "immutable"?

That's what I had at some point in the past, but it was both confusing and verbose. In many places you really want to abbreviate the meta-variable.

I'm tempted to try a different route: drop `var` and `const`, and rather do what we do with `null`, i.e., just have `mut?`, which would also match the text format.

dumblob:
`var` and `invar` as in [Dao](https://github.com/daokoder/dao/ )?

PR #331, opened 2022-11-01
Remove note about design being in flux

tlively:
Now that we're in phase 3, the design should be relatively stable.


titzer (approved):
lgtm!

PR #339, opened 2022-11-15
Add `final` attribute

rossberg:
This adds a spec of the final attribute for experimentation:

* In the abstract syntax, it is possible to write `sub $t* ...` or `sub final $t* ...`.
* In the binary format, `sub final` is encoded as `-0x32`, with the same format as `sub`.
* When validating a `sub` definition, none of the supertypes may be final.
* The non-sub shorthands (and therefor existing function types) are reinterpreted to `final`.

Due to the latter, a subtypable struct or array must now be written explicitly, with an empty list of supertypes:
```
(type (struct))  ;; final, no supertypes (shorthand)
(type (sub final (struct)))  ;; same in expanded form
(type (sub (struct)))  ;; non-final, no supertypes
(type (sub $t (struct)))  ;; non-final, with supertype
(type (sub final $t (struct)))  ;; final, with supertype
```

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -63,7 +63,7 @@ jobs:
       - run: pip install bikeshed && bikeshed update
       - run: pip install six
       - run: sudo apt-get update -y && sudo apt-get install -y latexmk texlive-latex-recommended texlive-latex-extra texlive-fonts-recommended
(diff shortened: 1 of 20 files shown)


rossberg:
Taking the liberty to land this, since I will already be out tomorrow, for 3 weeks.

PR #340, opened 2022-11-15
Add throwing instructions and funcref equality to Post-MVP

### LINKED ISSUE #208: Mapping Java null-pointer exceptions to WasmGC
Some more feedback from the J2CL team:

Many operations in Java, such as reading a member field of an object, throw a `NullPointerException` if the object is `null`. Translating such a member field access to a plain `struct.get` operation does not preserve Java semantics, because per the current proposal, `struct.get` traps if the object is `null`, and traps are not catchable in Wasm. (One might assume that NPE-throwing Java programs are buggy anyway and hence will not be executed in production; however unfortunately it turns out that such an assumption would be overly idealistic: real-world programs do rely on catching their own NPEs.) The obvious workaround is to put custom instruction sequences into the Wasm module, such as `br_on_null` before every `{struct,array}.{get,set}`. There is a concern that these custom checks bloat module size, and possibly hurt performance (depending on what exactly they'll look like, they might e.g. duplicate a check that the engine has to perform anyway -- though I expect that at least in some/simple cases, engines will be able to optimize away implicit checks when an explicit check has just been performed, i.e. `struct.get` should be able to not perform a null check if it follows a `br_on_null` instruction). The J2CL team is actively working on finalizing their implementation of such custom checks, which should give us some concrete numbers about their cost. I'll definitely follow up here once we have such data.

In the meantime, I already wanted to create awareness of this issue. Assuming the cost of custom checks ends up being painful, there are a few things we could do instead:

- we could make traps catchable, or introduce a new kind of catchable trap. (I expect that this idea will be unpopular; just mentioning it for completeness. There's no debate that uncatchable Wasm traps are a good way to represent C/C++ segfaults; however with the GC proposal we are clearly moving way beyond the conventions of C-like languages, and it may be worth re-evaluating this choice.)

- we could rebase the GC proposal on top of the exception handling proposal, and specify `struct.get` and friends to throw an exception instead of trapping. That would make WasmGC's own semantics a closer match to those managed languages that use NPEs or similar concepts, letting those languages piggyback more often on default Wasm behavior, instead of having to wrap so many operations in their own error-handling implementations.

- other ideas?

(Of course, we can also do nothing. As discussed many times before, it's impossible for WasmGC to be a perfect match for _every_ source language anyway, and maybe we shouldn't even try to match some source languages particularly well, so as to be source language agnostic. On the other hand, there is still the concern that WasmGC has to prove its usefulness in competitive comparison to existing, well-established and well-optimized technologies such as compiling to JS, so catering especially to certain popular use cases might just be the thing that makes the overall project viable and desirable.)

### LINKED ISSUE #239: Polymorphic devirtualization and funcref not being a subtype of eqref?
As mentioned in the last GC CG meeting, devirtualization helps quite a lot on j2cl, something like a [41% speedup](https://github.com/WebAssembly/meetings/blob/main/gc/2021/presentations/2021-08-24-zakai-optimization-update.pdf). That handles the case of a single call target being possible, that is, we load from a vtable and binaryen can infer that the vtable must contain a particular function reference, and so we replace the load with that reference, which then allows the call to become direct and even inlined.

Looking into the polymorphic case, that is, where there is a small number of possible function references but more than one, I was hoping to do something like this:
```wat
(struct.get vtable)

=>

(select
  (first possible function)
  (second possible function)
  (ref.eq (struct.get vtable) (first possible function))
)
```
Later optimizations can then replace a `call_ref` of a `select` of two function references into an if over two possible calls, etc. However, the condition of the select hits a problem, as funcref is not a subtype of eqref - function references cannot be compared for equality.

I couldn't find a detailed discussion of that, but IIRC the motivation was to allow VMs to optimize things like folding two identical functions into a single one, etc. That sounds reasonable, but the devirtualization issue shows that might be an optimization tradeoff which is not obvious?

Gathering some data, if I disable validation in binaryen then allowing 2 functions instead of 1 leads to 14K more places where we can turn a get from a vtable into a constant (well, a select over constants). Allowing 3 raises that to 17K, and 4 to 19K. (At some amount this becomes less useful, though, of course.) For comparison, the total number of `call_ref`s is 42K, so even with 2 functions we are talking about potentially optimizing away a third of indirect call sites, which sounds like it could be very significant.

Alternatives:
 * Rewrite the types, replacing the funcref with an `i32` index that we can select on. The problem is that we'd need to replace the vtable field in all relevant subtypes and supertypes, which may not be practical in general. Adding an additional field is another option, but would add memory and runtime overhead.
 * Use `ref.test` on the vtable. That might work if the different functions come from different types, which I believe is the general case (but I'd need to check). How fast is `ref.test` expected to be?
 * Perform such devirtualization in the VM and not the toolchain. Doing it statically is probably not reasonable (as a large LTO-style optimization), but using runtime profiling data it might be.

tlively:
Summarizing these post-MVP ideas in the documentation will allow us to close out their corresponding issues.

Closes #208.
Closes #239.

### DIFF
--- a/proposals/gc/Post-MVP.md
+++ b/proposals/gc/Post-MVP.md
@@ -21,6 +21,7 @@ See [overview](Overview.md) for addition background.
 * [Weak references](#weak-references)
 * [Method dispatch](#method-dispatch)
 * [Handle nondefaultable fields in `struct.new_default`](#handle-nondefaultable-fields-in-structnew_default)
+* [Throwing versions of trapping instructions](#throwing-versions-of-trapping-instructions)
 
 ## Bulk Operations
 
@@ -773,3 +774,18 @@ The problem could be addressed by extending the type system with features that a
 As proposed in #174. The idea is that `struct.new_default` would take values for nondefaultable fields in the struct from the stack. This would be a code size optimization over having to specify every single field with `struct.new` when allocating a struct containing nondefaultable fields.
 
 **Why Post-MVP:** This is a local code size optimization that does not affect expressivity, so getting it into the MVP is not urgent.
+
+
+## Throwing versions of trapping instructions
+
+Some language implementers have expressed interest in having throwing versions of operations like `struct.get` to simplify the implementation of e.g. Java's `NullPointerException`. It may be worth investigating what the potential code size or performance benefits of these instructions would be. See [#208](https://github.com/WebAssembly/gc/issues/208) for details and previous discussion.
+
+**Why Post-MVP:** Null checks followed by throws are easily expressible in the MVP, so there is no pressing need to add additional instructions for this pattern.
+
+
+## Equality-comparable funcref
+
+Being able to compare function references for equality would enable more precise polymorphic devirtualization. On the other hand, some polymorphic devirtualization should already be possible by taking advantage of patterns in vtables and letting functions be equality comparable would inhibit function deduplication optimizations. See [#239](https://github.com/WebAssembly/gc/issues/239) for details and previous discussion.
+
+**Why Post-MVP:** The benefits of this change are unknown and would likely be small, so it's not urgent that we get this into the MVP.
+
--- a/proposals/gc/Post-MVP.md
+++ b/proposals/gc/Post-MVP.md
@@ -780,7 +780,7 @@ As proposed in #174. The idea is that `struct.new_default` would take values for
 
 Some language implementers have expressed interest in having throwing versions of operations like `struct.get` to simplify the implementation of e.g. Java's `NullPointerException`. It may be worth investigating what the potential code size or performance benefits of these instructions would be. See [#208](https://github.com/WebAssembly/gc/issues/208) for details and previous discussion.
 
-**Why Post-MVP:** Null checks followed by throws are easily expressible in the MVP, so there is no pressing need to add additional instructions for this pattern.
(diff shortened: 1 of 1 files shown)


proposals/gc/Post-MVP.md
```diff
@@ -773,3 +774,18 @@ The problem could be addressed by extending the type system with features that a
 As proposed in #174. The idea is that `struct.new_default` would take values for nondefaultable fields in the struct from the stack. This would be a code size optimization over having to specify every single field with `struct.new` when allocating a struct containing nondefaultable fields.
 
 **Why Post-MVP:** This is a local code size optimization that does not affect expressivity, so getting it into the MVP is not urgent.
+
+
+## Throwing versions of trapping instructions
+
+Some language implementers have expressed interest in having throwing versions of operations like `struct.get` to simplify the implementation of e.g. Java's `NullPointerException`. It may be worth investigating what the potential code size or performance benefits of these instructions would be. See [#208](https://github.com/WebAssembly/gc/issues/208) for details and previous discussion.
+
+**Why Post-MVP:** Null checks followed by throws are easily expressible in the MVP, so there is no pressing need to add additional instructions for this pattern.
```
rossberg, line 783: I would add here that this feature is dependent on the exception proposal.

rossberg (approved):

Issue #342, opened 2022-12-09
Add br_on_cast[_fail] for non-nullable types


Liedtke:
**Current situation**:
`br_on_cast` operates on nullable input types. One can choose which path the `null` value "takes" by either using `br_on_cast` or `br_on_cast null`.

**Issue**:
For any given non-nullable input type, one of the two output types (either on the branch or on the fall-through case) will be nullable.
Ideally both output types (on branch / on non-branch) should be non-nullable for non-nullable inputs to not require useless `ref.as_non_null` casts afterwards.

**Proposal**:
This can be fixed by making `br_on_cast` and `br_on_cast_fail` polymorphic on the input type by changing
`br_on_cast $l null? ht : [t0* (ref null ht')] -> [t0* (ref null2? ht')]` to
`br_on_cast $l null? ht : [t0* (ref null4? ht')] -> [t0* (ref null2? ht')]`
with `null2` (and `null3` in the label type) not appearing if `null4` is is not there, independent of the presence of `null?`.
(The same applies to `br_on_cast_fail`).

The change doesn't really have any impact on the semantics of the instructions and would only affect the inferred result types.

**[Edit] Follow-up question**:
Should the same be applied for `ref.cast null?` for consistency?
That would mean that `ref.cast null` only produces a nullable type iff the input type is nullable.

jakobkummerow:
+1 to doing something here. It's unfortunate when perfectly non-nullable types go into a `br_*` instruction, and for no good reason a nullable result comes out (on one of the branches). Nullability (that presumably took some effort to rule out, at least in the shape of a `ref.as_non_null` instruction) shouldn't just re-materialize out of thin air.

rossberg:
Yes, good point, that's an oversight. Will fix.

rossberg:
Fixed by #343.

tlively:
To summarize, it seems like this current typing does the optimal thing when the input type is non-nullable or when the cast type is non-nullable, but there is room for improvement when both are nullable.

conrad-watt:
Just to note, the typing rules [prior to the change](https://github.com/WebAssembly/gc/blob/aa71d96412492c8a38130732dad5b866428889d7/proposals/gc/MVP.md#casts) had the desired property that @Liedtke's latest [comment](https://github.com/WebAssembly/gc/issues/342#issuecomment-1354505307) sketched, so presumably engines already implemented this.

[...]

PR #343, opened 2022-12-09
Relax nullability for br_on_cast

rossberg:
Address #342 and simplify rules on the way. Also adjust interpreter and add tests.

@Liedtke, PTAL.

### DIFF
--- a/proposals/gc/MVP.md
+++ b/proposals/gc/MVP.md
@@ -597,30 +597,30 @@ In particular, `ref.null` is typed as before, despite the introduction of `none`
 
 Casts work for both abstract and concrete types. In the latter case, they test if the operand's RTT is a sub-RTT of the target type.
 
-* `ref.test null? <heaptype>` checks whether a reference has a given heap type
-  - `ref.test rt : [(ref rt')] -> [i32]`
+* `ref.test <reftype>` tests whether a reference has a given type
+  - `ref.test rt : [rt'] -> [i32]`
     - iff `rt <: trt` and `rt' <: trt` for some `trt`
   - if `rt` contains `null`, returns 1 for null, otherwise 0
 
-* `ref.cast null? <heaptype>` tries to convert to a given heap type
-  - `ref.cast rt : [(ref rt')] -> [(ref rt)]`
+* `ref.cast <reftype>` tries to convert a reference to a given type
+  - `ref.cast rt : [rt'] -> [rt]`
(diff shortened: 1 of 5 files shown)


proposals/gc/MVP.md
```diff
@@ -598,35 +598,32 @@ In particular, `ref.null` is typed as before, despite the introduction of `none`
 Casts work for both abstract and concrete types. In the latter case, they test if the operand's RTT is a sub-RTT of the target type.
 
 * `ref.test null? <heaptype>` checks whether a reference has a given heap type
-  - `ref.test null? ht : [(ref null ht')] -> [i32]`
-    - iff `ht <: tht` and `ht' <: tht` where `tht` is a common super type
-  - if `null?` is present, returns 1 for null, otherwise 0
+  - `ref.test rt : [(ref rt')] -> [i32]`
+    - iff `rt <: trt` and `rt' <: trt` for some `trt`
+  - if `rt` contains `null`, returns 1 for null, otherwise 0
```
tlively, line 603: What does "`rt` contains `null`" mean? Is `rt` a reference type? Then it would be different in kind from `rt'`.

rossberg: Good point, I was using the wrong factorisation. Fixed.

proposals/gc/MVP.md
```diff
@@ -597,36 +597,33 @@ In particular, `ref.null` is typed as before, despite the introduction of `none`
...
+
+* `ref.cast <reftype>` tries to convert a reference to a given type
+  - `ref.cast rt : [rt'] -> [rt]`
+    - iff `rt <: trt` and `rt' <: trt` for some `trt`
   - traps if reference is not of requested type
-  - if `null?` is present, a null operand is passed through, otherwise traps on null
-  - equivalent to `(block $l (param anyref) (result (ref null? ht)) (br_on_cast null? ht $l) (unreachable))`
+  - if `rt` contains `null`, a null operand is passed through, otherwise traps on null
+  - equivalent to `(block $l (param anyref) (result rt) (br_on_cast $l rt) (unreachable))`
```
Liedtke, line 610: I think, this isn't true in all cases. For externref types or funcref types `anyref` would not be compatible.
is there a way to make it look nice while also applying to funcref types? Otherwise we could add a `(for input values of the anyref type hierarchy)` or a similar restriction to this line.

rossberg: Ah, indeed, this ceased to hold when we split the type hierarchy.

PR #348, opened 2023-01-20
Sync constant expression descriptions

takikawa:
This is a minor PR to sync the description of constant expressions in the proposal document (it's missing `extern.internalize/externalize`).

### DIFF
--- a/proposals/gc/MVP.md
+++ b/proposals/gc/MVP.md
@@ -655,6 +655,7 @@ In order to allow RTTs to be initialised as globals, the following extensions ar
 * `struct.new_canon` and `struct.new_canon_default` are constant instructions
 * `array.new_canon`, `array.new_canon_default`, and `array.new_canon_fixed` are constant instructions
   - Note: `array.new_canon_data` and `array.new_canon_elem` are not for the time being, see above
+* `extern.internalize` and `extern.externalize` are constant instructions
 * `global.get` is a constant instruction and can access preceding (immutable) global definitions, not just imports as in the MVP
 
 


tlively:
Sync from where? It doesn't look like they are constant instruction in V8 according to https://docs.google.com/document/d/14xWenN0lUoHa9TiDJd8yYR8ZKjgOhgO7iEmfrAuA1_M/

takikawa:
@tlively they're listed as constant in this section of the proposal explainer: https://github.com/WebAssembly/gc/blob/main/proposals/gc/MVP.md#external-conversion

>     extern.internalize converts an external value into the internal representation
>         extern.internalize : [(ref null1? extern)] -> [(ref null2? any)]
>             iff null1? = null2?
>         this is a constant instruction
>         note: this succeeds for all values, composing this with extern.externalize (in either order) yields the original value
> 
>     extern.externalize converts an internal value into the external representation
>         extern.externalize : [(ref null1? any)] -> [(ref null2? extern)]
>             iff null1? = null2?
>         this is a constant instruction
>         note: this succeeds for all values; moreover, composing this with extern.internalize (in either order) yields the original value

tlively:
Ah, sounds good. @jakobkummerow, does this sound good from the engine point of view?

jakobkummerow:
Making the spec text consistent definitely sounds good.

I have no particular implementation concerns. (That said, we're not currently implementing these two as constant, related to the point below.)

I would, however, ask the question of usefulness. What's a scenario where having these instructions available for initializing tables or globals is useful? So far we've been rather conservative in marking instructions as "constant", and only done so on an as-needed basis.

tlively:
I could imagine immutably importing a global of type externref and initializing another immutable global with the `extern.internalize` conversion of the imported global. Then the module internally uses the converted global and the external JS uses the imported global and the conversion cost is only paid once.

Issue #279, opened 2022-02-22
JS interop: no-frills approach


jakobkummerow:
This expands on @tlively 's idea here: https://github.com/WebAssembly/gc/issues/275#issuecomment-1043509538

I've been thinking about the situation created by multiple (semi-cooperative?) modules that use semantically-distinct types which get canonicalized per the isorecursive type system rules, but for which the involved modules would like to set up different JS interop behavior.

For a concrete example, suppose we have:
```wasm
; module A
(type $point (struct (field $x f64) (field $y f64)))
; module B
(type $size (struct (field $width f64) (field $height f64)))
```

If we imagine an "idiomatic" JS interop scheme, then that would likely mean some way of setting up named properties (so JS code can use `my_point.x` and `my_size.width`) and/or prototypes (`RTT_for_$size.prototype = { area() { return this.width * this.height; } }` etc). It is easy to see that this easily leads to collisions. In the "benign" case, we may simply end up with an object whose `.x` and `.width` properties are aliasing each other; but trouble arises when both types use the same name for different fields. If prototypes are assigned whole-sale (i.e. `.prototype = { area() {...} }`), they'll just clobber each other ("last one wins"); if they are assembled piecemeal (i.e. `.prototype.area = function() { ... }`), we'll again have problems if individual names clash; if we permit divergence between JS-exposed prototypes and Wasm-level typing (as some design explorations have suggested, e.g. by storing prototypes on RTTs, which would make these RTTs distinguishable in JS but indistinguishable in Wasm), then it seems unavoidable that this divergence would be observable from Wasm-module-embedding code and lead to some pretty weird behavior (such as prototypes "magically" changing or getting lost).

Whether the JS annotations are set up imperatively or declaratively doesn't make much of a difference for this problem.

One solution to these concerns is to stick our heads in the sand and hope that these cases won't happen much in practice. The primary mitigating factor is likely the fact that (at least with Java-like source languages) it is exceedingly unlikely for two semantically-different classes to have identically-typed vtables, and that alone prevents their isorecursive canonicalization. So maybe it's fine in practice.

Thomas' idea is the other solution. By _not_ exposing named fields or prototypes and instead fully relying on exported functions, we can avoid the whole problem. To be specific: the Wasm modules would, with some module-producer-defined naming convention, export free functions that act as accessors, e.g.:
```wasm
; module A
(func (export "$point_get_x") (param $p (ref $point)) (result f64) (struct.get (local.get 0) $point.$x))
; module B
[...]

PR #352, opened 2023-02-13
[spec] Initial JS API formal spec for the GC proposal

takikawa:
This PR adds formal JS API spec text based on the "no-frills" proposal in https://github.com/WebAssembly/gc/issues/279 and https://docs.google.com/document/d/17hCQXOyeSgogpJ0I0wir4LRmdvu4l7Oca6e1NkbVN8M/.

I've also opened an issue to discuss certain aspects of this PR: https://github.com/WebAssembly/gc/issues/353

### DIFF
--- a/document/js-api/index.bs
+++ b/document/js-api/index.bs
@@ -90,6 +90,7 @@ urlPrefix: https://tc39.github.io/ecma262/; spec: ECMASCRIPT
         text: IterableToList; url: sec-iterabletolist
         text: ToBigInt64; url: #sec-tobigint64
         text: BigInt; url: #sec-ecmascript-language-types-bigint-type
+        text: MakeBasicObject; url: #sec-makebasicobject
     type: abstract-op
         text: CreateMethodProperty; url: sec-createmethodproperty
 urlPrefix: https://webassembly.github.io/spec/core/; spec: WebAssembly; type: dfn
@@ -113,6 +114,10 @@ urlPrefix: https://webassembly.github.io/spec/core/; spec: WebAssembly; type: df
         text: ref.null
         text: ref.func
         text: ref.extern
+        <!-- FIXME: correct these links once the GC proposal formal spec is complete -->
+        text: ref.i31
+        text: ref.array
+        text: ref.struct
     text: function index; url: syntax/modules.html#syntax-funcidx
     text: function instance; url: exec/runtime.html#function-instances
     text: store_init; url: appendix/embedding.html#embed-store-init
@@ -142,6 +147,12 @@ urlPrefix: https://webassembly.github.io/spec/core/; spec: WebAssembly; type: df
     text: global_read; url: appendix/embedding.html#embed-global-read
     text: global_write; url: appendix/embedding.html#embed-global-write
     text: error; url: appendix/embedding.html#embed-error
+    <!-- FIXME: these links should be updated with the GC proposal formal spec -->
+    text: i31_new; url: appendix/embedding.html#embed-gc
+    text: i31_get; url: appendix/embedding.html#embed-gc
+    text: ref_cast; url: appendix/embedding.html#embed-gc
+    text: extern_internalize; url: appendix/embedding.html#embed-gc
+    text: extern_externalize; url: appendix/embedding.html#embed-gc
     text: store; url: exec/runtime.html#syntax-store
     text: table type; url: syntax/types.html#syntax-tabletype
     text: table address; url: exec/runtime.html#syntax-tableaddr
@@ -149,6 +160,8 @@ urlPrefix: https://webassembly.github.io/spec/core/; spec: WebAssembly; type: df
     text: memory address; url: exec/runtime.html#syntax-memaddr
     text: global address; url: exec/runtime.html#syntax-globaladdr
     text: extern address; url: exec/runtime.html#syntax-externaddr
+    <!-- FIXME: Update this link with the GC proposal formal spec -->
+    text: object address; url: exec/runtime.html#syntax-objectaddr
     url: syntax/types.html#syntax-numtype
         text: i32
         text: i64
@@ -159,6 +172,12 @@ urlPrefix: https://webassembly.github.io/spec/core/; spec: WebAssembly; type: df
         text: reftype
         text: funcref
         text: externref
+        text: ref
+    url: syntax/types.html#heap-types; for: heap-type
+        text: extern
+        text: func
+        text: i31
+        text: any
     text: function element; url: exec/runtime.html#syntax-funcelem
     text: import component; url: syntax/modules.html#imports
     text: external value; url: exec/runtime.html#syntax-externval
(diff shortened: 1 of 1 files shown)


document/js-api/index.bs
```diff
@@ -1120,26 +1144,155 @@ The algorithm <dfn>ToWebAssemblyValue</dfn>(|v|, |type|) coerces a JavaScript va
 1. If |type| is [=f64=],
     1. Let |f64| be ? [=ToNumber=](|v|).
     1. Return [=f64.const=] |f64|.
-1. If |type| is [=funcref=],
+1. If |type| is of the form [=ref=] |null| |heaptype| and is a subtype of [=ref=] |null| [=heap-type/func=],
```
rossberg, line 1147: It looks like you may be able to avoid some duplication by first checking the null case, and only then distinguish between func/extern/any cases.

document/js-api/index.bs
```diff
@@ -1120,26 +1144,155 @@ The algorithm <dfn>ToWebAssemblyValue</dfn>(|v|, |type|) coerces a JavaScript va
...
     1. If |v| is null,
-        1. Return [=ref.null=] [=funcref=].
+        1. If |null| is present,
+            1. Return [=ref.null=] |heaptype|.
+        1. Otherwise,
+            1. Throw a {{TypeError}}.
     1. If |v| is an [=Exported Function=],
         1. Let |funcaddr| be the value of |v|'s \[[FunctionAddress]] internal slot.
-        1. Return [=ref.func=] |funcaddr|.
+        1. Let |casted| be [=ref_cast=](|heaptype|, |funcaddr|).
```
rossberg, line 1155: Casts do not change the value, so if the cast succeeds, it would simply return the original funcaddr. But the cast may also fail, so the result type of this operator would have to include an error case.

Hence it is probably more natural for the embedding API to provide a predicate like
```
ref_typecheck(addr, heaptype) : bool
```
instead. (Perhaps we even want to separate getting the ref's own type from a subtype function on two types.)

document/js-api/index.bs
```diff
@@ -1120,26 +1144,155 @@ The algorithm <dfn>ToWebAssemblyValue</dfn>(|v|, |type|) coerces a JavaScript va
...
     1. Let [=extern address=] |externaddr| be the smallest address such that |map|[|externaddr|] [=map/exists=] is false.
     1. [=map/Set=] |map|[|externaddr|] to |v|.
     1. Return [=ref.extern=] |externaddr|.
+1. If |type| is of the form [=ref=] |null| |heaptype| and is a subtype of [=ref=] |null| [=any=],
+    1. If |v| is null,
+        1. If |null| is present,
+            1. Return [=ref.null=] |heaptype|.
+        1. Otherwise,
+            1. Throw a {{TypeError}}.
+    1. If |heaptype| is [=i31=],
```
rossberg, line 1176: This condition seems wrong: i31's need to be generated for type `any` or `eq` as well, and the conversion should be the same (and accept the same values) regardless of the type. Likewise, you'll need to allow arbitrary JS values for `any`, not just ints or GC objects.

All non-null values ought to be converted uniformly by first performing internalize and then a type check. The internalize function will then have to do the job to convert numbers to i31 if they are within range, and treat them as host values otherwise.

This indeed implies that the JS spec needs to amend the core spec by customising the meaning of internalize/externalize. The core spec cannot really define it, because it doesn't know what host values are. We'll need some hook in the core spec for this host-specific semantics.

document/js-api/index.bs
```diff
@@ -1090,6 +1109,11 @@ The algorithm <dfn>ToJSValue</dfn>(|w|) coerces a [=WebAssembly value=] to a Jav
 1. If |w| is of the form [=ref.null=] <var ignore>t</var>, return null.
 1. If |w| is of the form [=ref.func=] |funcaddr|, return the result of creating [=a new Exported Function=] from |funcaddr|.
 1. If |w| is of the form [=ref.extern=] |externaddr|, return the result of [=retrieving an extern value=] from |externaddr|.
+1. If |w| is of the form [=ref.array=] |arrayaddr|, return the result of creating [=a new GC Exported Object=] from |arrayaddr|.
+1. If |w| is of the form [=ref.struct=] |structaddr|, return the result of creating [=a new GC Exported Object=] from |structaddr|.
+1. If |w| is of the form [=ref.i31=] |i31addr|,
+    1. Let |i32| be [=i31_get=](|i31addr|).
+    1. Return [=the Number value=] for |i32|.
```
rossberg, line 1116: I believe that externalize needs to be invoke here somehow. In particular, that will also handle values that are not Wasm GC values, such as internalised JS values. The core spec will have to introduce the notion of host object for those.

document/js-api/index.bs
takikawa: Thanks for catching that, I wonder if it would make sense to call the operation `ref_test` to mirror the `ref.test` instruction that is similar in functionality.

document/js-api/index.bs
takikawa: Thanks, fixed in the latest commit.

document/js-api/index.bs
takikawa: > I think all non-null values ought to be converted uniformly by first performing internalize and then a type check. 

I've changed it to use to use this approach now: https://github.com/WebAssembly/gc/commit/2d39b2254b8198b49b8e4a03a3de6fcfca18bfe0#diff-73846d0b7899a94fe050d3c0ba447da607bebedf80f8b5ed31cb4d164756f048R1170

Also added a new section in the spec that details the requirements on externalize & internalize: https://github.com/WebAssembly/gc/pull/352/commits/2d39b2254b8198b49b8e4a03a3de6fcfca18bfe0#diff-73846d0b7899a94fe050d3c0ba447da607bebedf80f8b5ed31cb4d164756f048R1348

document/js-api/index.bs
takikawa: The latest commit uses `ref_test` for now (not tied to this naming, could use something else or go with a different way of checking) and should check the failure case.

rossberg (approved):
LGTM. I have some more thoughts about simplifying the ex/internalize story such that it does not actually require patching the core semantics, but that can be adjusted once the core spec is written.

document/js-api/index.bs
```diff
@@ -1120,26 +1147,142 @@ The algorithm <dfn>ToWebAssemblyValue</dfn>(|v|, |type|) coerces a JavaScript va
...
+
+An [=Exported GC Object=] contains an \[[ObjectAddress]] internal slot, which holds a [=object address=] relative to the [=surrounding agent=]'s [=associated store=],
+and an \[[ObjectKind]] internal slot, which holds the string value "struct" or "array".
+
+The internal methods of an [=Exported GC Object=] use the following implementations.
+
+<div algorithm>
+  The <dfn>\[[GetPrototypeOf]] internal method of an Exported GC Object</dfn> <var ignore>O</var> takes no arguments and returns null. It performs the following steps when called:
+
+    1. Return null.
```
takikawa, line 1197: I updated [[GetPrototypeOf]] to return null to match the updated google doc, also see this comment: https://github.com/WebAssembly/gc/pull/355#discussion_r1111711131

takikawa:
@rossberg Thanks for the review on this earlier, is it ok to merge this already? We discussed this in the GC subgroup meeting today and nobody mentioned any blockers to merging (and we can adjust to match the core spec later, as you said).

rossberg:
Sure, I believe I already approved the PR.

takikawa:
> Sure, I believe I already approved the PR.

Ok thanks! I don't think I have merge access so if you or someone else with merge access (@tlively maybe?) could do it, that'd be great.

PR #355, opened 2023-02-17
[test] Add tentative JS API tests for Exported GC Object

takikawa:
Adds tests for the "Exported GC Object" described in the proposed JS API spec in https://github.com/WebAssembly/gc/pull/352.

This requires updating the wasm module builder helper library, which is now fairly de-synced with the V8 library it's based on. I've only synced a bare minimum in this PR to be able to construct some GC modules (syncing the whole thing causes other tests to fail).

The V8 version also uses a slightly different opcode numbering than the current spec document in this repo. In this PR I went with the spec numbering (in the end they will all need to be synchronized).

This doesn't have tests for all the ToWasmValue/ToJSValue cases for conversions, which would also be useful to add.

### DIFF
--- a/test/js-api/gc/exported-object.tentative.any.js
+++ b/test/js-api/gc/exported-object.tentative.any.js
@@ -50,14 +50,10 @@ test(() => {
 test(() => {
   const struct = functions.makeStruct();
   const array = functions.makeArray();
-  struct.foo = 5;
-  assert_equals(struct.foo, undefined);
-  array.foo = 5;
-  assert_equals(array.foo, undefined);
-  struct[0] = 5;
-  assert_equals(struct[0], undefined);
-  array[0] = 5;
-  assert_equals(array[0], undefined);
+  assert_throws_js(TypeError, () => { struct.foo = 5; });
+  assert_throws_js(TypeError, () => { array.foo = 5; });
+  assert_throws_js(TypeError, () => { struct[0] = 5; });
+  assert_throws_js(TypeError, () => { array[0] = 5; });
 }, "property assignment (non-strict mode)");
 
 test(() => {
@@ -87,14 +83,10 @@ test(() => {
 test(() => {
   const struct = functions.makeStruct();
   const array = functions.makeArray();
-  const structNamesLength = Object.getOwnPropertyNames(struct).length;
-  const arrayNamesLength = Object.getOwnPropertyNames(array).length;
-  delete struct.foo;
-  delete struct[0];
-  delete array.foo;
-  delete array[0];
-  assert_equals(Object.getOwnPropertyNames(struct), structNamesLength);
-  assert_equals(Object.getOwnPropertyNames(array), arrayNamesLength);
+  assert_throws_js(TypeError, () => delete struct.foo);
+  assert_throws_js(TypeError, () => delete struct[0]);
+  assert_throws_js(TypeError, () => delete array.foo);
+  assert_throws_js(TypeError, () => delete array[0]);
 }, "delete (non-strict mode)");
 
 test(() => {
--- a/test/js-api/gc/exported-object.tentative.any.js
+++ b/test/js-api/gc/exported-object.tentative.any.js
@@ -72,8 +72,8 @@ test(() => {
 test(() => {
   const struct = functions.makeStruct()
   const array = functions.makeArray()
-  assert_throws_js(TypeError, () => Object.getPrototypeOf(struct));
-  assert_throws_js(TypeError, () => Object.getPrototypeOf(array));
+  assert_equals(Object.getPrototypeOf(struct), null);
+  assert_equals(Object.getPrototypeOf(array), null);
 }, "getPrototypeOf");
 
 test(() => {
--- a/test/js-api/gc/exported-object.tentative.any.js
+++ b/test/js-api/gc/exported-object.tentative.any.js
@@ -1,0 +1,160 @@
+// META: global=window,dedicatedworker,jsshell
+// META: script=/wasm/jsapi/wasm-module-builder.js
+
+let functions = {};
+setup(() => {
+  const builder = new WasmModuleBuilder();
+
+  const structIndex = builder.addStruct([makeField(kWasmI32, true)]);
+  const arrayIndex = builder.addArray(kWasmI32, true);
+  const structRef = wasmRefType(structIndex);
+  const arrayRef = wasmRefType(arrayIndex);
+
+  builder
+    .addFunction("makeStruct", makeSig_r_v(structRef))
+    .addBody([...wasmI32Const(42),
+              ...GCInstr(kExprStructNew), structIndex])
+    .exportFunc();
+
+  builder
+    .addFunction("makeArray", makeSig_r_v(arrayRef))
+    .addBody([...wasmI32Const(5), ...wasmI32Const(42),
+              ...GCInstr(kExprArrayNew), arrayIndex])
+    .exportFunc();
+
+  const buffer = builder.toBuffer()
+  const module = new WebAssembly.Module(buffer);
+  const instance = new WebAssembly.Instance(module, {});
+  functions = instance.exports;
+});
+
+test(() => {
+  const struct = functions.makeStruct()
+  const array = functions.makeArray()
+  assert_equals(struct.foo, undefined);
+  assert_equals(struct[0], undefined);
+  assert_equals(array.foo, undefined);
+  assert_equals(array[0], undefined);
+}, "property access");
+
+test(() => {
+  const struct = functions.makeStruct()
+  const array = functions.makeArray()
+  assert_throws_js(TypeError, () => { struct.foo = 5; });
+  assert_throws_js(TypeError, () => { array.foo = 5; });
+  assert_throws_js(TypeError, () => { struct[0] = 5; });
+  assert_throws_js(TypeError, () => { array[0] = 5; });
+}, "property assignment");
+
+test(() => {
+  const struct = functions.makeStruct()
+  const array = functions.makeArray()
+  assert_equals(Object.getOwnPropertyNames(struct).length, 0);
+  assert_equals(Object.getOwnPropertyNames(array).length, 0);
+}, "ownPropertyNames");
+
+test(() => {
+  const struct = functions.makeStruct()
+  const array = functions.makeArray()
+  assert_throws_js(TypeError, () => Object.defineProperty(struct, "foo", { value: 1 }));
+  assert_throws_js(TypeError, () => Object.defineProperty(array, "foo", { value: 1 }));
+}, "defineProperty");
+
+test(() => {
+  const struct = functions.makeStruct()
+  const array = functions.makeArray()
+  assert_throws_js(TypeError, () => delete struct.foo);
+  assert_throws_js(TypeError, () => delete struct[0]);
+  assert_throws_js(TypeError, () => delete array.foo);
+  assert_throws_js(TypeError, () => delete array[0]);
+}, "delete");
+
+test(() => {
+  const struct = functions.makeStruct()
+  const array = functions.makeArray()
+  assert_throws_js(TypeError, () => Object.getPrototypeOf(struct));
+  assert_throws_js(TypeError, () => Object.getPrototypeOf(array));
+}, "getPrototypeOf");
+
+test(() => {
+  const struct = functions.makeStruct()
+  const array = functions.makeArray()
+  assert_throws_js(TypeError, () => Object.setPrototypeOf(struct, {}));
+  assert_throws_js(TypeError, () => Object.setPrototypeOf(array, {}));
+}, "setPrototypeOf");
+
+test(() => {
+  const struct = functions.makeStruct()
+  const array = functions.makeArray()
+  assert_false(Object.isExtensible(struct));
+  assert_false(Object.isExtensible(array));
+}, "isExtensible");
+
+test(() => {
+  const struct = functions.makeStruct()
+  const array = functions.makeArray()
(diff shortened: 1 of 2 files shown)


test/js-api/wasm-module-builder.js
```diff
@@ -374,10 +390,48 @@ let kExprRefIsNull = 0xd1;
...
 let kAtomicPrefix = 0xfe;
 
+// Use these for multi-byte instructions (opcode > 0x7F needing two LEB bytes):
+function GCInstr(opcode) {
+  if (opcode <= 0x7F) return [kGCPrefix, opcode];
+  return [kGCPrefix, 0x80 | (opcode & 0x7F), opcode >> 7];
+}
+
+// GC opcodes
+let kExprStructNew = 0x01;
```
takikawa, line 405: Note: the opcode numbers for `struct.new` and `array.new` differ from V8/SM currently.

test/js-api/gc/exported-object.tentative.any.js
```diff
@@ -0,0 +1,160 @@
...
+  const array = functions.makeArray()
+  assert_equals(typeof struct, "object");
+  assert_equals(typeof array, "object");
+}, "typeof");
+
+test(() => {
+  const struct = functions.makeStruct()
+  const array = functions.makeArray()
+  assert_throws_js(TypeError, () => struct.toString());
+  assert_equals(Object.prototype.toString.call(struct), "[object Object]");
```
takikawa, line 118: These `toString.call` tests failed in V8. I'm guessing it might be because `[[Get]]` was originally going to be specified to throw, but was changed to return `undefined` and that changes the `toString` behavior.

test/js-api/gc/exported-object.tentative.any.js
```diff
@@ -0,0 +1,160 @@
...
+  assert_throws_js(TypeError, () => delete struct.foo);
+  assert_throws_js(TypeError, () => delete struct[0]);
+  assert_throws_js(TypeError, () => delete array.foo);
+  assert_throws_js(TypeError, () => delete array[0]);
+}, "delete");
+
+test(() => {
+  const struct = functions.makeStruct()
+  const array = functions.makeArray()
+  assert_throws_js(TypeError, () => Object.getPrototypeOf(struct));
```
takikawa, line 75: When I tried this in V8, this test failed because the `getPrototypeOf` didn't throw.

takikawa:
Other than the points I mentioned in the tests above, I was able to run this in V8 by manually importing it into `v8/test/wasm-js/tests` and editing a few things like opcodes. I also ran it against a WIP implementation in JSC.

test/js-api/gc/exported-object.tentative.any.js
```diff
@@ -0,0 +1,160 @@
...
+  assert_equals(struct.foo, undefined);
+  assert_equals(struct[0], undefined);
+  assert_equals(array.foo, undefined);
+  assert_equals(array[0], undefined);
+}, "property access");
+
+test(() => {
+  const struct = functions.makeStruct()
+  const array = functions.makeArray()
+  assert_throws_js(TypeError, () => { struct.foo = 5; });
```
takikawa, line 43: Right now [[Set]] is specified to always throw, which matches the test cases here.

It is arguably a bit inconsistent with the rest of JS though, because usually [[Set]] returns false on failure but doesn't throw. In the usual case, that means that in non-strict mode, `struct.foo = 5` will not add/mutate the field, but will silently fail. With Wasm structs/arrays, it will fail even in non-strict mode. Not sure anyone will care about this difference, but I thought it might be worth mentioning.

test/js-api/gc/exported-object.tentative.any.js
jakobkummerow: Yeah, when we made [[Get]] not throw we also had to make [[GetPrototypeOf]] not throw (but return `null` instead). Looks like I forgot to update the draft doc accordingly, sorry about that. Fixed it there just now.

test/js-api/gc/exported-object.tentative.any.js
jakobkummerow: Yes, I believe this implementation is a leftover from when [[Get]] was throwing, and we'll have to change that in V8.

(FWIW, `struct.toString()` should continue to throw, just like `Object.create(null).toString()`.)

Liedtke: Yes, indeed, that was done due to `[[Get]]` throwing. I reverted that now: https://chromium-review.googlesource.com/c/v8/v8/+/4271370.

test/js-api/gc/exported-object.tentative.any.js
takikawa: Ok thanks, I updated this test case to check for null instead as well. I think this actually makes it easier for JSC to implement [[Get]]'s behavior too, though I'm wondering if it'll be sufficient to future-proof for shared structs.

takikawa:
Who would be the best person to review this and provide review approval? @Ms2ger would you be interested in taking a look?

ajklein:
I can't seem to add him as a reviewer, but it would make sense for @syg to do at least an unofficial review of this given the intersection with [JS Structs](https://github.com/tc39/proposal-structs).

test/js-api/gc/exported-object.tentative.any.js
Ms2ger: Yeah, returning false probably makes more sense. In any case, we should probably make sure to test these cases both in strict and loose mode

test/js-api/gc/exported-object.tentative.any.js
```diff
@@ -0,0 +1,160 @@
...
+  assert_throws_js(TypeError, () => Object.defineProperty(array, "foo", { value: 1 }));
+}, "defineProperty");
+
+test(() => {
+  const struct = functions.makeStruct()
+  const array = functions.makeArray()
+  assert_throws_js(TypeError, () => delete struct.foo);
+  assert_throws_js(TypeError, () => delete struct[0]);
+  assert_throws_js(TypeError, () => delete array.foo);
+  assert_throws_js(TypeError, () => delete array[0]);
```
Ms2ger, line 69: Same comment about testing in strict and loose mode (and probably changing the spec)

test/js-api/gc/exported-object.tentative.any.js
```diff
@@ -0,0 +1,160 @@
...
+
+test(() => {
+  const struct = functions.makeStruct()
+  const array = functions.makeArray()
+  const map = new WeakMap();
+  map.set(struct, "struct");
+  map.set(array, "array");
+  assert_equals(map.get(struct), "struct");
+  assert_equals(map.get(array), "array");
+}, "GC objects as weak map keys");
```
Ms2ger, line 160: Do you want to check `WeakSet` to complete the square?

takikawa: Added

test/js-api/gc/exported-object.tentative.any.js
takikawa: Done

test/js-api/gc/exported-object.tentative.any.js
takikawa: Yeah, that makes sense. I've changed the test now to have strict and non-strict versions, with no error raised in the non-strict version (and the property shouldn't be added/deleted). We'll have to change the spec to match this though.

test/js-api/wasm-module-builder.js
```diff
@@ -374,10 +390,48 @@ let kExprRefIsNull = 0xd1;
...
 // Prefix opcodes
+let kGCPrefix = 0xfb;
 let kNumericPrefix = 0xfc;
 let kSimdPrefix = 0xfd;
 let kAtomicPrefix = 0xfe;
 
+// Use these for multi-byte instructions (opcode > 0x7F needing two LEB bytes):
+function GCInstr(opcode) {
+  if (opcode <= 0x7F) return [kGCPrefix, opcode];
+  return [kGCPrefix, 0x80 | (opcode & 0x7F), opcode >> 7];
```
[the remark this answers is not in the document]
takikawa: It might be able to.

In general for these `wasm-module-builder.js` changes I took the upstream V8 code as-is when I could, so merging new changes would be easier without a complicated diff. Do you prefer that I make changes here anyway and let it diverge with upstream? (also applies to other comments)

test/js-api/wasm-module-builder.js
```diff
@@ -728,10 +826,24 @@ class WasmModuleBuilder {
     this.explicit.push(this.createCustomSection(name, bytes));
   }
 
-  addType(type) {
-    this.types.push(type);
-    var pl = type.params.length;  // should have params
-    var rl = type.results.length; // should have results
+  // We use {is_final = true} so that the MVP syntax is generated for
+  // signatures.
```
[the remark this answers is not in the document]
takikawa: In the GC proposal type declarations for structs & arrays can be declared final or not. Since the finality is declared for particular type defs I think it has to be type-by-type.

takikawa:
Thanks for the comments, I resolved them aside from the ones about the module builder (see comment). The main change is to allow not throwing an exception in non-strict mode (and silently failing) for adding properties and deleting.

This modified version should succeed on V8 except for the non-strict tests. On JSC it also passes (except non-strict) with some minor changes in opcodes. In JSC I implemented a quick fix for the non-strict case and got all tests to pass too.

test/js-api/wasm-module-builder.js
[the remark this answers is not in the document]
Ms2ger: Perhaps someone from the V8 side could comment. @jakobkummerow seems to have some blame around here.

Would be a nice project for someone with time to consolidate all the versions going around.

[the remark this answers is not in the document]
jakobkummerow: Yeah, this probably could use `wasmSignedLeb`, but it also doesn't matter as it seems exceedingly unlikely that we'll have that many opcodes any time soon (if ever). From a V8 perspective, we see our `wasm-module-builder.js` as very frequently changing code anyway: we update it whenever we need to (typically to support some new feature), and that happens a lot. Basically the reason we have our own thing there rather than importing someone else's implementation is because we need to change it so much, and want that to have low friction and low latency.

If you want to submit a patch to improve it, I'd volunteer to review that, but I'm currently not inclined to spend my own time to make this code more polished-looking.

jakobkummerow:
I have a concern about the change to not throw in non-strict mode.
I think this new behavior (not throwing) would make perfect sense if we aimed to specify some final state of affairs. But the idea behind the throwing was specifically to keep the door wide open to future changes in behavior here (when, post-MVP, we discuss options for richer interop stories).

That's why I'd prefer to keep throwing in both strict and non-strict mode for now.
*If* in the future we reach the conclusion that the "no-frills" interop is actually enough, and we want to declare it final, then we can relax the non-strict mode behavior at that point.
Whereas if we develop any approach that actually allows `foo.bar` notation in the future, it could very well be very important for not breaking the web with that rollout that we can be sure that nobody, not even non-strict mode users, is currently shipping any code that would break with this rollout.

Also, in general, when a PR named "add tentative tests" makes such a change to behavior, I think this needs to be explicitly called out to a wider audience (one might even argue that it should happen as a separate PR).

test/js-api/gc/exported-object.tentative.any.js
littledan: Yeah, I'm a little worried about future-proofing for improved prototypes later (related to alignment with shared structs). Is this is the only way we can make it so that async functions can return Wasm GC objects without blowing up? If so, it seems like this is the best possible compromise.

rossberg:
What is the status of this PR?

rossberg:
@takikawa, ping, is this PR still relevant?

takikawa:
> @takikawa, ping, is this PR still relevant?

Hi @rossberg, yes I think it's relevant but it needs a rebase to adjust for the final opcodes and to address Jakob's concerns. I have some local changes that I was planning to push after testing.

takikawa:
Had to force push due to the rebase. This should now be updated to throw in non-strict mode as well, like @jakobkummerow requested (i.e., reverted to the original test behavior for this).

It should also be updated to match the final opcodes in MVP.md. I've tested it on JSC/WebKit but haven't had a chance to run it on V8 or SpiderMonkey.

jakobkummerow:
@takikawa For testing this on V8 with the final binary encoding, you can patch in [crrev.com/c/4756846](https://chromium-review.googlesource.com/c/v8/v8/+/4756846). We're planning to land that after the Phase 4 vote.

takikawa:
I've now tested this on V8 (on commit 014e0bb5ac4691c85b1da2dac579af77d02d17b7, already updated to new opcodes) in the `wasm-js` test suite and on Firefox (commit f4b0e304ecebf431616ce014dd198c1d361b37c1, with patches from [bug 1845374](https://bugzilla.mozilla.org/show_bug.cgi?id=1845374) applied) with WPT tests.

I also pushed some fixes that I found while testing this, all in the `wasm-module-builder.js` helper.

On Firefox, some exception handling proposal tests do fail. This is because the exception tests (that are in WPT as tentative tests) use some type code names that have changed in the upstream `wasm-module-builder.js`. In this case, I think the exception tests (specifically [this one](https://github.com/WebAssembly/exception-handling/blob/main/test/js-api/exception/basic.tentative.any.js#L85)) should just be changed to match as it's in phase 3.

takikawa:
In any case, I think this is ready to merge if other people are happy with it. Separately, I'm planning to work on on the rest of the JS API tests (testing `ToWebAssemblyValue` and `ToJSValue`) soon.

PR #359, opened 2023-03-22
Add 2nd type annotations on casts

### LINKED ISSUE #342: Add br_on_cast[_fail] for non-nullable types

rossberg:
This specifies the changes to cast instructions as discussed, and implements them in interpreter and test suite. Also removes _canon suffixes.

Two points of note:

1. For the binary format, I tentatively went with a single opcode (0xfb4f) for all the variants of cast branches, followed by a flag byte currently carrying 3 bits: one for each nullability, and one for direction (_fail vs regular). This avoids allocating 8 separate opcodes (or 6, since 2 wouldn't validate) and also allows for later extensions without combinatorial explosion of opcodes. Makes more sense to me, but let's discuss if you think different.
2. I tweaked the text format to use regular reference type notation on cast instructions (e.g., `ref.cast (ref null $t)`, `br_on_cast $l anyref (ref $t)`). This looks less weird, especially with the two annotations on branches, and it also conveniently allows using the shorthands.

Fixes #342 and supersedes PR #347.


jakobkummerow:
`0xfb4f` is currently unused, so that's a good choice for a binary encoding that doesn't break ongoing test coverage for the various prototype implementations. (And of course we'll compact the encoding space once the proposal's instruction set is final.)

tlively (commented):
I would personally prefer to allocate the 6 opcodes rather than have a flags field for the branching casts because a 5-byte best case encoding for those instructions is already extremely unfortunate and I would like to avoid pushing that to 6 bytes.

titzer:
I think having taken/not taken (_fail vs regular) as separate opcodes would be helpful. I am also ok with encoding the null-handling as an immediate. I am also fine with 6 different opcodes.

proposals/gc/MVP.md
```diff
@@ -609,31 +609,29 @@ Casts work for both abstract and concrete types. In the latter case, they test i
...
-    - iff `$l : [t0* t']`
-    - and `rt <: t'`
-    - and `rt <: trt` and `rt' <: trt` for some `trt`
+* `br_on_cast <labelidx> <reftype> <reftype>` branches if a reference has a given type
+  - `br_on_cast $l rt1 rt2 : [t0* rt1] -> [t0* rt1]`
+    - iff `$l : [t0* rt2]`
+    - and `rt2 <: rt1`
   - passes operand along with branch under target type, plus possible extra args
-  - if `rt` contains `null`, branches on null, otherwise does not
+  - if `rt2` contains `null`, branches on null, otherwise does not
```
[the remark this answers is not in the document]
rossberg: Oops, yes, you are right.

rossberg:
I still expect these to remain fairly cold instructions, so I'm not too worried about the byte. My bigger concern with 6 opcodes is that it doesn't scale well. If we add another dimension (say, RTTs ;) ) it already doubles to 12, and so on.

The other oddity is that by making 2 combinations syntactically unrepresentable, it's technically not two reftype immediates anymore, but something rather irregular.

rossberg:
@conrad-watt, I'm not sure I follow. This has nothing to do with the operational behaviour. What I'm saying is that, although the abstract syntax is essentially `br_on_cast rt1 rt2`, there would in fact be some combinations of types that prohibited _syntactically_, and not merely by the static type constraint `rt2 <: rt1` like all others. That is a phasing conflation.

interpreter/valid/valid.ml
```diff
@@ -444,41 +444,37 @@ let rec check_instr (c : context) (e : instr) (s : infer_result_type) : infer_in
...
+      ("type mismatch: instruction requires type " ^ string_of_ref_type rt2 ^
        " but label has " ^ string_of_result_type (label c x));
     let ts0, t1 = Lib.List.split_last (label c x) in
-    require (match_val_type c.types (RefT rt) t1) e.at
-      ("type mismatch: instruction requires type " ^ string_of_ref_type rt ^
+    require (match_val_type c.types (RefT rt2) t1) e.at
+      ("type mismatch: instruction requires type " ^ string_of_ref_type rt2 ^
        " but label has " ^ string_of_result_type (label c x));
-    (ts0 @ [RefT rt']) --> (ts0 @ [RefT rt']), []
+    (ts0 @ [RefT rt1]) --> (ts0 @ [RefT rt1]), []
```
conrad-watt, line 461: Is this refining the output type as intended? I would have expected something other than `rt1` as the output, depending on whether `rt2` contains null.

conrad-watt:
I guess I'm trying to make a broader argument that it makes more sense for us to think, abstractly, of the operations being split up as follows:

`br_on_cast lab null? ht ht`
`br_on_cast_fail lab null? ht ht`
`br_on_cast_nonnull lab ht ht`
`br_on_cast_nonnull_fail lab ht ht`

This is essentially what I sketched [here](https://github.com/WebAssembly/gc/issues/342#issuecomment-1421284394) but with an explicit second heap type annotation. My intuition is that this makes the semantic behaviour (dependent on the presence or absence of `null?`) and the refinement of the branch/non-branch output type easier to reason about.

I think if one instead rolls all of these cases inside the `null?` parts of two `reftype` annotations, one ends up with a bunch of side-conditions to distinguish each case (and refinement) that don't feel totally natural to me - i.e. the `reftype` annotations aren't really used as `reftypes`, but more like "heaptype with a helpful bool flag" (see also my review comment just above).

proposals/gc/MVP.md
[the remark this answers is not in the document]
rossberg: Fixed, here and below. Also added tests for these cases.

rossberg:
I have to disagree. This would be overcomplicating matters. The semantics is actually fairly simple and regular now:

- Dynamically, the instruction simply checks subtyping between two reference types: the one of the operand and the target type.

- Statically, it has an input type and two output types:
  - the output type of the success path is the annotated target type,
  - the output type of the failure path is the _diff_ of input type and target type.

If we had more general union types, that diff would literally be a set-like diff. With our current algebra, null is the only element we can express individually, so the diff is approximate and only affects that one.

interpreter/valid/valid.ml
rossberg: Yes, implemented now. I had simply forgotten the type refinement entirely. :)

tlively:
Changes LGTM. FWIW I had previously viewed the reftype annotations as "heap types with helpful bools" as Conrad put it, but with two of them corresponding to the input and desired output, I actually think of them as reftypes now. I guess the weird part is that they're not encoded in the binary format using the normal reftype encoding, but I definitely wouldn't want the encoding to be any larger, so I'm glad we're not doing that. I don't feel too strongly about which of the current options we go with.

conrad-watt (approved):
I don't feel too strongly either and the diff approach in the validation algorithm is neat, so LGTM

Liedtke:
There are a few questions / concerns that came up while trying to add the new instruction to V8:

1) *One opcode*: While I don't really mind having all the `br_on_cast` logic behind a single op code, I feel tempted to say that it's a bit unfortunate that we have one opcode in the binary format but 2 operations (`br_on_cast` + `br_on_cast_fail`) in the text format. Assuming that the branch depth immediate is invalid and we wanted to create an error `br_on_cast[0]: invalid branch depth '10'` [^1], we wouldn't know when creating the error whether the instruction is a `br_on_cast` or a `br_on_cast_fail` because this is expressed after the branch depth immediate. So it might be confusing that `br_on_cast` and `-fail` are technically the same operation / opcode even though they look completely different in any disassembler output / source code.
2) *Cast from non-nullable to nullable*: As we have two null bits, it is possible to specify that the source type is non-nullable but the target type is nullable. In that case the validation type for the variant of a successful cast becomes nullable, correct? (Also meaning that this flag combination is not treated as a validation error.)
3) When casting from a nullable type to a nullable type using `br_on_cast`, the fall-through type becomes non-nullable, so the decoder refines the type to something more specific. However, this refinement only applies to the nullability, not to the heap type? Asked differently, when doing an upcast in a `br_on_cast_fail` e.g. from `arrayref` to `anyref`, the branch is de-facto unreachable as the cast can never fail. I assume that the branch signature still has to align with the source type as specified in the `br_on_cast_fail` immediates? (So the branch label has to be typed as `arrayref` or any supertype in this example.)

[^1]: Funnily we don't print the opcode name for invalid branch depth immediates but we do it for invalid type immediates, so this is just an inconsistency in our (current) error messages and doesn't affect the argument.

rossberg:
Re (2): As you can see from the typing rule in the PR, it simply requires that for `br_on_cast rt1 rt2`, the second type is a subtype of the first -- because it's a downcast. From the subtyping rules for reference types it follows immediately that this will reject the case `br_on_cast (ref $t1) (ref null $t2)`, because the latter type never is a subtype of the former.

Re (3): Yes, the refinement only applies to nullability, since our type system isn't expressive enough to express any form of refinement on the heap type. Computing a meaningful type difference for such a refinement would only work for union types, which we do not have. Consequently, the type difference operation is an approximation that leaves the heap type unchanged.

I don't follow your question regarding an upcast. You neither can nor need to use br_on_cast for upcasts -- going from arrayref to anyref is just subsumption (implicit use of subtyping), so never requires a cast, nor could it ever fail.

proposals/gc/MVP.md
```diff
@@ -609,31 +609,33 @@ Casts work for both abstract and concrete types. In the latter case, they test i
 
 * `ref.test <reftype>` tests whether a reference has a given type
   - `ref.test rt : [rt'] -> [i32]`
-    - iff `rt <: trt` and `rt' <: trt` for some `trt`
+    - iff `rt <: rt'`
```
eqrion, line 612: Does this change mean that:
```
(func
  (param arrayref)
  local.get 0
  ref.test structref
  drop
)
```

is not valid because `structref </: arrayref`? Which seems to contradict that downcasts among sibling hierarchies are allowed (written later in this doc).

rossberg: Don't worry, it's just a simplification of the rule that I realised is possible and that should have no effect on what's allowed. Basically, it was redundant to distinguish between `rt'` and `trt` here before, that was only necessary for the branching rules. Your example is fine, because the rule can always pick `anyref` for `rt'`, which would be the principal type.

eqrion: Ah, yes that makes sense. Thanks.

Liedtke:
@rossberg Thanks for clarifying! I didn't notice the semantic change that the target type has to be a subtype of the source type immediate as previously the type on the value stack only had to be in the same type hierarchy.
I wrongly assumed "If I use the type on the value stack as the source type immediate (`rt1`), there isn't any difference to the old behavior".

rossberg:
Right, with the old behaviour, the situation was more complicated, because we needed to allow sufficient flexibility in the typing rule to maintain type preservation under reduction. With an explicit type annotation, it all is much simpler.

By ordinary subtyping, the actual stack operand can of course still be a subtype of the input annotation -- and therefor a super-, sub-, or even sibling type of the target. So it is no less expressive than before.

rossberg:
@eqrion, yes, I'd be fine with making the distinction via some separate token if others feel strongly about it.

tlively:
I would also prefer that we use a single instruction name if we are going to use a single opcode. This will improve error messages in V8 (without complicating their existing code) as well.

tlively:
We decided at the meeting this morning to use two different opcodes for the fail and non-fail cases.

rossberg:
Done.

tlively (approved):
New encoding LGTM.

tlively:
@Liedtke and @jakobkummerow, will we be able to implement these new opcodes (0x4e and 0x4f for br_on_cast and br_on_cast_fail) directly? Binaryen hadn't yet implemented the previous encoding, so this will not be a breaking change for any Binaryen users.

Liedtke:
@tlively Yes, because binaryen didn't implement it and this PR hasn't been merged yet, we went ahead and implemented the new opcode and updated decoding for `0x4f` without a fallback today. (If required we could allow the old fail bit in the flags byte to "switch" the opcode from `0x4f` to `0x4e`.)

PR #360, opened 2023-03-22
Document new implementation-defined limits

### LINKED ISSUE #335: Implementation-defined limits
The JS spec has a section on [implementation-defined limits](https://webassembly.github.io/spec/js-api/index.html#limits). Several new limits will have to be introduced for the GC proposal.

1. The limit on the number of types is currently one million. We should define a limits on the number of rec groups and the size of each rec group. Should we reinterpret the current limit as the limit on rec groups or should we additionally enforce a limit on the total number of types across all rec groups? Should we put the total number of types into the binary format of the type section so engines can continue to check this limit eagerly?

2. We should define a limit on the number of fields allowed in a struct.

3. We should define a limit on the number of arguments `array.new_fixed` can take (keeping in mind that this will be used to initialize very large global arrays assuming https://github.com/WebAssembly/gc/issues/260 is not solved for the MVP).

4. We should define runtime limits on the size of arrays.

And I am probably missing additional limits as well.

tlively:
Based on discussion in #335 and subgroup meetings.

Closes #335.

### DIFF
--- a/proposals/gc/MVP.md
+++ b/proposals/gc/MVP.md
@@ -779,19 +779,9 @@ The opcode for heap types is encoded as an `s33`.
 | 0xfb70 | `extern.internalize` | |
 | 0xfb71 | `extern.externalize` | |
 
-## Implementation-defined Limits
-
-The following limits will be added to the Implementation-defined Limits [section](https://webassembly.github.io/spec/js-api/index.html#limits) of the JS API.
-
- - The maximum number of recursion groups is 1000000. (The maximum number of individual types remains unchanged and is also 1000000.)
- - The maximum number of struct fields is 10000.
- - The maximum number of operands to `array.new_fixed` is 10000.
- - The maximum length of a supertype chain is 63. (A type declared with no supertypes has a supertype chain of length 0)
-
 ## JS API
 
-See [GC JS API document](MVP-JS.md) .
-
+See [GC JS API document](../../document/js-api/index.bs).
 
 ## Questions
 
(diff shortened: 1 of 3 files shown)


tlively:
* main
  * **PR #360** <a href="https://app.graphite.dev/github/pr/WebAssembly/gc/360" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/WebAssembly/gc/360?utm_source=stack-comment).

tlively:
The one limit we discussed that is not in this PR is the byte limit on the size of global initializers. I decided not to include it because it is under discussion separately in the extended-const proposal: https://github.com/WebAssembly/extended-const/issues/15.

proposals/gc/MVP.md
```diff
@@ -779,6 +779,14 @@ The opcode for heap types is encoded as an `s33`.
 | 0xfb70 | `extern.internalize` | |
 | 0xfb71 | `extern.externalize` | |
 
+## Implementation-defined Limits
```
rossberg, line 782: Since these are constraints of the JS API spec, it should make more sense to move this section down below and turn it into a subsection of that one.

tlively:
I've now fixed the link in MVP.md to point to the actual bikeshed document containing the JS spec (rather than the outdated MVP-JS.md) and I've put the new limits directly into the relevant section in the bikeshed doc.

rossberg:
Hm, I think we should update the MVP-JS doc to match the status quo. Otherwise, we're losing the documentation of that part of the proposal and its design. Looking at the final spec provides no useful information about what the proposal was about. We are intentionally keeping around design docs in the proposals directory, after all.

tlively:
Ok, this now documents the limits in a revamped MVP-JS.md doc as well as in the actual spec doc.

rossberg (approved):

Issue #370, opened 2023-05-02
Final WasmGC opcodes


tlively:
One of the last things we need to do is finalize the opcodes for all the new WasmGC instructions. @titzer has filed #337 proposing a nano-optimization for the type encodings, but we have not discussed the final instruction encodings yet.

Would anyone like to propose a concrete final encoding scheme?

tlively:
At the meeting today, @rossberg offered to come up with a proposal for the final opcodes. No one shared any opinions about how the opcodes should be laid out.

tlively:
Closing this because #372 has landed.

PR #363, opened 2023-03-27
initial bulk array op implementations

conrad-watt:
See https://github.com/WebAssembly/gc/issues/313, and more specifically [this comment](https://github.com/WebAssembly/gc/issues/313#issuecomment-1453013131).

This patch _only_ updates the interpreter, the overview is not updated.

Note - this patch currently using the following opcode encodings in the hope of provoking more permanent suggestions:
```
| ArrayCopy (x, y)     -> op 0xfb; op 0x18; var x; var y
| ArrayFill x          -> op 0xfb; op 0x0f; var x
| ArrayInitData (x, y) -> op 0xfb; op 0x52; var x; var y
| ArrayInitElem (x, y) -> op 0xfb; op 0x53; var x; var y
```

PR #372, opened 2023-05-09
Final opcodes

### LINKED ISSUE #337: Binary encoding of value type codes with heap type indices
We haven't chosen the final binary encoding for value type codes introduced in this proposal.

One issue is that decoding value types is a very hot operation in in-place interpretation, particularly the declaration of local variables as well as block types (though block types are just decoded and ignored). Until this proposal, all value types are decodable as a single byte (or single LEB--we've been inconsistent on that so far). With this proposal, some value type codes have a heaptype immediate. Thus an in-place interpreter can no longer simply decode (or skip) a single LEB.

I've resorted to doing a table lookup to see if the value type code indicates that a heaptype index should follow (`ref` or `ref null`). But with one tweak to the assignment of type codes, it's possible to turn that into a bit-test.

```
11111111 I32 = -1
11111110 I64 = -2
11111101 F32 = -3
11111100 F64 = -4
11111011 V128 = -5
11110000 FUNCREF = -16
11101111 EXTERNREF = -17
11101110 ANYREF = -18
11101101 EQREF = -19
11101100 REF_NULL = -20
11101011 REF = -21
11101010 I31REF = -22
11101001 NULLFUNCREF = -23
11101000 NULLEXTERNREF = -24
11100111 STRUCTREF = -25
11100110 ARRAYREF = -26
11100101 NULLREF = -27
11000000 EmptyBlock = -64
```

It turns out that we have encoding space to make the 6th bit the determinant as to whether a heaptype immediate follows. We just need to make `ref` and `ref null` < `-32`. That makes the 6th bit `0` in case of `ref`, `ref null`; empty block still needs a separate check. (Perhaps it is still possible to rearrange so that only one test is necessary).

That constrains the "short encoding" space to 32 entries if we want to preserve the bit test property in future additions.

rossberg:
Here is my proposal for the final opcode reordering (fixes #337 and #370).

Trying to group related instructions together and leave space for future extensions in respective opcode ranges.

Also reordering type opcodes to group nullary constructors together and before non-nullary ones (c.f. #337).

### DIFF
--- a/proposals/gc/MVP.md
+++ b/proposals/gc/MVP.md
@@ -674,25 +674,25 @@ This extends the [encodings](https://github.com/WebAssembly/function-references/
 
 | Opcode | Type            |
 | ------ | --------------- |
-| -0x06  | `i8`            |
-| -0x07  | `i16`           |
+| -0x08  | `i8`            |
+| -0x09  | `i16`           |
 
 #### Reference Types
 
 | Opcode | Type            | Parameters | Note |
 | ------ | --------------- | ---------- | ---- |
+| -0x0d  | `nullfuncref`   |            | shorthand |
+| -0x0e  | `nullexternref` |            | shorthand |
+| -0x0f  | `nullref`       |            | shorthand |
 | -0x10  | `funcref`       |            | shorthand, from reftype proposal |
 | -0x11  | `externref`     |            | shorthand, from reftype proposal |
 | -0x12  | `anyref`        |            | shorthand |
 | -0x13  | `eqref`         |            | shorthand |
-| -0x14  | `(ref null ht)` | `ht : heaptype (s33)` | from funcref proposal |
-| -0x15  | `(ref ht)`      | `ht : heaptype (s33)` | from funcref proposal |
-| -0x16  | `i31ref`        |            | shorthand |
-| -0x17  | `nullfuncref`   |            | shorthand |
-| -0x18  | `nullexternref` |            | shorthand |
-| -0x19  | `structref`     |            | shorthand |
-| -0x1a  | `arrayref`      |            | shorthand |
-| -0x1b  | `nullref`       |            | shorthand |
+| -0x14  | `i31ref`        |            | shorthand |
+| -0x15  | `structref`     |            | shorthand |
+| -0x16  | `arrayref`      |            | shorthand |
+| -0x1c  | `(ref ht)`      | `ht : heaptype (s33)` | from funcref proposal |
+| -0x1d  | `(ref null ht)` | `ht : heaptype (s33)` | from funcref proposal |
 
 #### Heap Types
 
@@ -701,42 +701,45 @@ The opcode for heap types is encoded as an `s33`.
 | Opcode | Type            | Parameters | Note |
 | ------ | --------------- | ---------- | ---- |
 | i >= 0 | `(type i)`      |            | from funcref proposal |
+| -0x0d  | `nofunc`        |            | |
+| -0x0e  | `noextern`      |            | |
+| -0x0f  | `none`          |            | |
 | -0x10  | `func`          |            | from funcref proposal |
 | -0x11  | `extern`        |            | from funcref proposal |
 | -0x12  | `any`           |            | |
 | -0x13  | `eq`            |            | |
-| -0x16  | `i31`           |            | |
-| -0x17  | `nofunc`        |            | |
-| -0x18  | `noextern`      |            | |
-| -0x19  | `struct`        |            | |
-| -0x1a  | `array`         |            | |
-| -0x1b  | `none`          |            | |
+| -0x14  | `i31`           |            | |
+| -0x15  | `struct`        |            | |
+| -0x16  | `array`         |            | |
 
 #### Structured Types
 
-| Opcode | Type            | Parameters |
-| ------ | --------------- | ---------- |
-| -0x21  | `struct ft*`    | `ft* : vec(fieldtype)` |
-| -0x22  | `array ft`      | `ft : fieldtype`       |
+| Opcode | Type            | Parameters | Note |
+| ------ | --------------- | ---------- | ---- |
+| -0x20  | `func t1* t2*`  | `t1* : vec(valtype)`, `t2* : vec(valtype)` | from Wasm 1.0 |
+| -0x21  | `struct ft*`    | `ft* : vec(fieldtype)` | |
+| -0x22  | `array ft`      | `ft : fieldtype`       | |
 
 #### Subtypes
 
 | Opcode | Type            | Parameters | Note |
 | ------ | --------------- | ---------- | ---- |
+| -0x20  | `func t1* t2*`  | `t1* : vec(valtype)`, `t2* : vec(valtype)` | shorthand |
 | -0x21  | `struct ft*`    | `ft* : vec(fieldtype)` | shorthand |
 | -0x22  | `array ft`      | `ft : fieldtype`       | shorthand |
 | -0x30  | `sub $t* st`    | `$t* : vec(typeidx)`, `st : strtype` | |
-| -0x32  | `sub final $t* st` | `$t* : vec(typeidx)`, `st : strtype` | |
+| -0x31  | `sub final $t* st` | `$t* : vec(typeidx)`, `st : strtype` | |
 
 #### Defined Types
 
 | Opcode | Type            | Parameters | Note |
 | ------ | --------------- | ---------- | ---- |
+| -0x20  | `func t1* t2*`  | `t1* : vec(valtype)`, `t2* : vec(valtype)` | shorthand |
 | -0x21  | `struct ft*`    | `ft* : vec(fieldtype)` | shorthand |
 | -0x22  | `array ft`      | `ft : fieldtype`       | shorthand |
 | -0x30  | `sub $t* st`    | `$t* : vec(typeidx)`, `st : strtype` | shorthand |
-| -0x31  | `rec dt*`       | `dt* : vec(subtype)` | |
-| -0x32  | `sub final $t* st` | `$t* : vec(typeidx)`, `st : strtype` | shorthand |
+| -0x31  | `sub final $t* st` | `$t* : vec(typeidx)`, `st : strtype` | shorthand |
+| -0x34  | `rec dt*`       | `dt* : vec(subtype)` | |
 
 #### Field Types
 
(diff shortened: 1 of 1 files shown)


jakobkummerow:
Looks OK.

The bulk array instructions (from #363) appear to be missing.

Personally, I would have compacted the opcode encoding space a lot more, essentially to assign a contiguous range. My reasoning is that whatever gaps we choose to leave now have an exceedingly small likelihood to turn out to have appropriate size, so as we add instructions in future proposals we'll likely end up with a wildly fragmented mix of remaining gaps that were too large and others that were too small where we then had to simply append new opcodes at the end of the range or squeeze them into "non-fitting" gaps. OTOH, it's not a big issue because (1) at the end of the day we're just choosing between different kinds of ugly for the encoding space a couple of proposals down the line, and (2) prettiness or ugliness of the binary encoding doesn't actually matter.

tlively:
FWIW, we tried to leave useful gaps in the SIMD proposal opcode space, and they did not turn out to be useful at all. If we had instead used a compact opcode space, we might not have needed to use an extra LEB byte for the relaxed SIMD opcodes.

I don't feel strongly either way, though.

titzer:
+1 to just compacting the opcode space without gaps. The 1-byte opcode space is important, but IMO there is marginal benefit, if any, in leaving holes in the prefixed opcode spaces.

askeksa-google:
Nit: I'd suggest to keep different encodings for the same instruction together, i.e. put `ref.test (ref null ht)` at 0xfb41and `ref.cast (ref ht)` at 0xfb42.

rossberg:
I agree that we should fill up gaps before going to two bytes, but currently we are far away from hitting that boundary, and never may. Is there any particular advantage in compacting before that? I thought that leaving some gaps actually worked reasonably well for some of the Wasm 1.0 code space.

@askeksa-google, good point, fixed.

jakobkummerow:
Is there any particular advantage to leaving gaps?

FWIW, here's the summary:
This PR currently lists 27 `0xfb`-prefixed instructions (still missing 4 array bulk instructions), and pads them with 53 gaps. These gaps are located in the following ranges of opcodes:
- 6 in "struct allocators"
- 4 in "struct accessors"
- 11 in "array allocators" (4 of which might be used for the bulk instructions)
- 11 in "array accessors" (4 of which might be used for the bulk instructions)
- 5 in "i31 instructions"
- 6 in "extern conversions"
- 4 in "cast instructions", non-branching group
- 6 in "cast instructions", branching group (assuming `0x4f` is seen as the end of the range)

One data point to illustrate the low likelihood of gap sizes being lucky guesses: Aske has previously suggested a set of 8 branching i31 instructions, which (if we end up deciding to adopt them in a future proposal) wouldn't have an obvious fit in any of these gaps.

titzer:
I don't think we should leave gaps in the prefixed opcode spaces. First, as mentioned above, I think it's unlikely we'll predict the size of gaps properly, meaning there will likely always be vestigial gaps. Second, leaving gaps reduces the number of two-byte encodings being utilized; in the future, will new instructions start filling in gaps if they'd otherwise overflow into the 3 byte space? Third, engines may way to use some of the unused opcode space for internal things, like quickened versions of bytecodes, which benefits interpreters.

rossberg:
Coming back to this:

* We left gaps in the MVP opcodes, and that proved relatively useful, both for new types and for new control instructions.
* The proposed opcodes do not just leave gaps, they also provide some additional prefixing hints. For example, 
  * 0xfb0X struct instructions
  * 0xfb1X/0xfb2X array instructions
  * 0xfb3X other reference instructions
  * 0xfb4X cast instructions
  I occasionally found such a thing useful in the past when I had to stare at binary code.
* I happily believe this approach failed for SIMD, but that's because SIMD is extremely large and irregular.
* I'm not sure I see the disadvantages. Note that we still have plenty of room in the single-byte space. We can still start filling up gaps or flow over into other prefixes once the need arises.

@titzer, I didn't follow your point about custom opcodes. Isn't that orthogonal?

titzer:
> @titzer, I didn't follow your point about custom opcodes. Isn't that orthogonal?

It's a minor point, but it's often useful to have a few opcodes left over for the VM, in each of the opcode spaces. This is easier to do if the opcode space fills up from the lowest numbers first. It doesn't work very well to try to use an opcode that is currently a hole and later filled up.

> We left gaps in the MVP opcodes, and that proved relatively useful, both for new types and for new control instructions.

The 1-byte opcode space is a much more important space, but in retrospect, I think we would have been fine to have not left holes in it.

rossberg:
> This is easier to do if the opcode space fills up from the lowest numbers first.

I see, but this is mostly a question of leaving sufficient space at the end, right? Which I would think we have even now.

tlively:
Let's discuss this at the subgroup meeting tomorrow. If we don't come to an agreement through discussion, I propose that we take a popularity vote to settle the issue. It doesn't seem important enough to be worth spending much additional energy or time on.

titzer:
I won't be able to attend tomorrow because of a conflict, but I would generally prefer that we pack the prefixed opcode spaces rather than leaving holes. Is a wider CG discussion warranted, as this could be a potentially precedent-setting decision?

askeksa-google:
If the purpose is to have some nice groupings, we could still do that while leaving significantly smaller holes, e.g.:
- `0x00+`: Struct instructions
- `0x10+`: Array instructions (including bulk)
- `0x20+`: Conversions (extern conversions from `0x24`)
- `0x28+`: Casts

tlively:
> Is a wider CG discussion warranted, as this could be a potentially precedent-setting decision?

Maybe, but it seems extremely low priority and CG time is hard to come by these days.

---

We didn't have a critical mass of people to resolve this at the meeting yesterday, but I'd like to take the temperature here. We are only adding 31 prefixed opcodes, so even if we leave generous holes, we will not be close to overflowing to the second LEB byte, which would require 128 opcodes. Given that, do you:

🚀 : prefer leaving holes
❤️ : prefer not leaving holes
👀 : not care at all

tlively:
As discussed at the last subgroup meeting, I've pushed a commit that simply removes all of the holes in the instruction opcode space but preserves the order of instructions.

Here's a sheet to help visualize the difference: https://docs.google.com/spreadsheets/d/1eaMdpL-MwLOwWgj-mWgv_tNq5crP_nemQ-PW54ss-Q8/edit?usp=sharing

proposals/gc/MVP.md
```diff
@@ -761,31 +761,35 @@ The opcode for heap types is encoded as an `s33`.
...
-| 0xfb32 | `i31.get_u` |  |
-| 0xfb38 | `extern.internalize` | | extern conversions (0x38+) |
-| 0xfb39 | `extern.externalize` | |
-| 0xfb40 | `ref.test (ref ht)` | `ht : heaptype` | cast instructions (0x40+) |
-| 0xfb41 | `ref.test (ref null ht)` | `ht : heaptype` |
-| 0xfb42 | `ref.cast (ref ht)` | `ht : heaptype` |
-| 0xfb43 | `ref.cast (ref null ht)` | `ht : heaptype` |
-| 0xfb48 | `br_on_cast $l (ref null1? ht1) (ref null2? ht2)` | `flags : u8`, $l : labelidx`, `ht1 : heaptype`, `ht2 : heaptype` |
-| 0xfb49 | `br_on_cast_fail $l (ref null1? ht1) (ref null2? ht2)` | `flags : u8`, $l : labelidx`, `ht1 : heaptype`, `ht2 : heaptype` |
+| 0xfb02 | `struct.get $t i` | `$t : typeidx`, `i : fieldidx` | struct accessors (0x08+) |
```
rossberg, line 811: The notes about group prefixes are no longer accurate.

proposals/gc/MVP.md
```diff
@@ -761,31 +761,35 @@ The opcode for heap types is encoded as an `s33`.
...
+| 0xfb14 | `i31.new` | i31 conversions (0x30+) |
+| 0xfb15 | `i31.get_s` |  |
+| 0xfb16 | `i31.get_u` |  |
+| 0xfb17 | `extern.internalize` | | extern conversions (0x38+) |
+| 0xfb18 | `extern.externalize` | |
+| 0xfb19 | `ref.test (ref ht)` | `ht : heaptype` | cast instructions (0x40+) |
+| 0xfb1a | `ref.test (ref null ht)` | `ht : heaptype` |
+| 0xfb1b | `ref.cast (ref ht)` | `ht : heaptype` |
+| 0xfb1c | `ref.cast (ref null ht)` | `ht : heaptype` |
+| 0xfb1d | `br_on_cast $l (ref null1? ht1) (ref null2? ht2)` | `flags : u8`, $l : labelidx`, `ht1 : heaptype`, `ht2 : heaptype` |
```
rossberg, line 838:
```suggestion
| 0xfb1d | `br_on_cast $l (ref null1? ht1) (ref null2? ht2)` | `flags : u8`,` $l : labelidx`, `ht1 : heaptype`, `ht2 : heaptype` |
```

proposals/gc/MVP.md
```diff
@@ -761,31 +761,35 @@ The opcode for heap types is encoded as an `s33`.
...
+| 0xfb1e | `br_on_cast_fail $l (ref null1? ht1) (ref null2? ht2)` | `flags : u8`, $l : labelidx`, `ht1 : heaptype`, `ht2 : heaptype` |
```
rossberg, line 839:
```suggestion
| 0xfb1e | `br_on_cast_fail $l (ref null1? ht1) (ref null2? ht2)` | `flags : u8`, `$l : labelidx`, `ht1 : heaptype`, `ht2 : heaptype` |
```

proposals/gc/MVP.md
```diff
@@ -748,42 +748,45 @@ The opcode for heap types is encoded as an `s33`.
...
 | Opcode | Type            | Parameters | Note |
 | ------ | --------------- | ---------- | ---- |
+| -0x20  | `func t1* t2*`  | `t1* : vec(valtype)`, `t2* : vec(valtype)` | shorthand |
 | -0x21  | `struct ft*`    | `ft* : vec(fieldtype)` | shorthand |
 | -0x22  | `array ft`      | `ft : fieldtype`       | shorthand |
 | -0x30  | `sub $t* st`    | `$t* : vec(typeidx)`, `st : strtype` | shorthand |
-| -0x31  | `rec dt*`       | `dt* : vec(subtype)` | |
-| -0x32  | `sub final $t* st` | `$t* : vec(typeidx)`, `st : strtype` | shorthand |
+| -0x31  | `sub final $t* st` | `$t* : vec(typeidx)`, `st : strtype` | shorthand |
+| -0x34  | `rec dt*`       | `dt* : vec(subtype)` | |
```
bashor, line 789: It looks like the interpreter wasn't changed accordingly

rossberg: Oops, thanks. Fixed. Also changed the rec opcode to -0x32, not sure why I picked something else before.

askeksa-google:
I do think it's worth skipping an opcode to align the pairs, especially since for `ref.test` and `ref.cast`, the lsb then becomes a null flag with the same meaning as the null flags in `br_on_cast[_fail]`.

We still stay within the first 32 opcodes.

tlively:
Another way we could align the pairs would be to take the three i31 operations and move them to the end. @rossberg, would you prefer to have a single hole before the i31 operations or to move the i31 operations to the end?

rossberg:
I don't care much either way, but if you change anything, please sync the interpreter as well in this PR.

askeksa-google:
If we move the `i31` instructions to the end, I suggest we also move the `extern` instructions after the casts, in order to have all "conversions" (`extern` and `i31`) grouped together.

interpreter/binary/decode.ml
```diff
@@ -610,24 +610,24 @@ let rec instr s =
...
-    | 0x1al -> ref_test (Null, heap_type s)
-    | 0x1bl -> ref_cast (NoNull, heap_type s)
-    | 0x1cl -> ref_cast (Null, heap_type s)
-    | 0x1dl | 0x1el as opcode ->
+    | 0x1cl -> i31_new
+    | 0x1dl -> i31_get_s
+    | 0x1el -> i31_get_u
+
+    | 0x1al -> extern_internalize
+    | 0x1bl -> extern_externalize
```
rossberg, line 618: Nit: Please reorder cases in this match so that they appear in order of their value.

rossberg:
FWIW, if we care for bits that way, shouldn't we also make it so that any _u/_s pairs of instructions only differ by LSB?

tlively:
Perhaps, although I would say that `struct.get{,_u,_s}` and `array.get(,_u,_s}` are triples, so it doesn't matter so much there, leaving only `i31.get_u` and `i31.get_s`. @askeksa-google, do you have an opinion here?

rossberg:
I notice that this is already not the case for some MVP instructions, so I think we can leave it as is.

tlively:
Cool, let's put this into something of a "final comment period." If you are ok with the currently proposed encoding, please give this comment a 🚀 react. Otherwise, please give it an 👀 react and comment with a specific change you would like to see. After a week or so, if we have consensus, we can go ahead and merge this.

PR #373, opened 2023-05-09
Spec abstract syntax for GC types and instructions

rossberg:
Does not yet include type definitions.

### DIFF
--- a/document/core/syntax/instructions.rst
+++ b/document/core/syntax/instructions.rst
@@ -528,10 +528,10 @@ Similarly, :ref:`arrays <syntax-type-array>` can be allocated either with an exp
 Furthermore, |ARRAYNEWFIXED| allocates an array with statically fixed size,
 and |ARRAYNEWDATA| and |ARRAYNEWELEM| allocate an array and initialize it from a :ref:`data <syntax-data>` or :ref:`element <syntax-elem>` segement, respectively.
 The remaining array instructions access individual slots,
-again allowing for different sign extension modes in the case of :ref:`packed <syntax-type-packed>` a storage type.
+again allowing for different sign extension modes in the case of a :ref:`packed <syntax-type-packed>` storage type.
 Last, |ARRAYLEN| produces the length of an array.
 
-The instructions |I31NEW|, |I31GETS|, and |I31GETU| convert between type |I31| and an unboxed :ref:`scalars <syntax-i31>`.
+The instructions |I31NEW|, |I31GETS|, and |I31GETU| convert between type |I31| and an unboxed :ref:`scalar <syntax-i31>`.
 
 The instructions |EXTERNINTERNALIZE| and |EXTERNEXTERNALIZE| allow lossless conversion between references represented as type :math:`(\REF~\NULL~\EXTERN)`| and as :math:`(\REF~\NULL~\ANY)`.
 
--- a/document/core/syntax/instructions.rst
+++ b/document/core/syntax/instructions.rst
@@ -415,12 +415,15 @@ Occasionally, it is convenient to group operators together according to the foll
    \end{array}
 
 
-.. index:: ! reference instruction, reference, null
+.. index:: ! reference instruction, reference, null, cast, heap type, reference type
    pair: abstract syntax; instruction
 .. _syntax-ref.null:
 .. _syntax-ref.func:
 .. _syntax-ref.is_null:
 .. _syntax-ref.as_non_null:
+.. _syntax-ref.eq:
+.. _syntax-ref.test:
+.. _syntax-ref.cast:
 .. _syntax-instr-ref:
 
 Reference Instructions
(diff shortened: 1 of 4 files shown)


tlively:
Is there a good way to view the rendered versions of spec PRs like this?

document/core/syntax/instructions.rst
```diff
@@ -435,11 +438,102 @@ Instructions in this group are concerned with accessing :ref:`references <syntax
...
+
+The instructions |STRUCTNEW| and |STRUCTNEWDEFAULT| allocate a new :ref:`structure <syntax-type-struct>`, initializing them either with operands or with default values.
+The remaining instructions on structs access individual fields,
+allowing for different sign extension modes in the case of :ref:`packed <syntax-type-packed>` storage types.
+
+Similarly, :ref:`arrays <syntax-type-array>` can be allocated either with an explicit initialization operand or a default value.
+Furthermore, |ARRAYNEWFIXED| allocates an array with statically fixed size,
+and |ARRAYNEWDATA| and |ARRAYNEWELEM| allocate an array and initialize it from a :ref:`data <syntax-data>` or :ref:`element <syntax-elem>` segement, respectively.
+The remaining array instructions access individual slots,
+again allowing for different sign extension modes in the case of :ref:`packed <syntax-type-packed>` a storage type.
```
tlively, line 531: Should the "a" be before the "packed" here?

document/core/syntax/instructions.rst
```diff
@@ -435,11 +438,102 @@ Instructions in this group are concerned with accessing :ref:`references <syntax
...
+Last, |ARRAYLEN| produces the length of an array.
+
+The instructions |I31NEW|, |I31GETS|, and |I31GETU| convert between type |I31| and an unboxed :ref:`scalars <syntax-i31>`.
```
tlively, line 534:
```suggestion
The instructions |I31NEW|, |I31GETS|, and |I31GETU| convert between type |I31| and an unboxed :ref:`scalar <syntax-i31>`.
```

document/core/syntax/types.rst
```diff
@@ -73,39 +73,84 @@ Conventions
...
+     \ANY ~|~ \EQ ~|~ \I31 ~|~ \STRUCT ~|~ \ARRAY ~|~ \NONE \\&&|&
+     \typeidx ~|~ \deftype ~|~ \BOT \\
    \end{array}
 
-The type |FUNC| denotes the infinite union of all types of :ref:`functions <syntax-func>`, regardless of their concrete :ref:`function types <syntax-functype>`.
+A heap type is either *abstract* or *concrete*.
+
+The abstract type |FUNC| denotes the infinite union (i.e., supertype) of all :ref:`function types <syntax-functype>`, regardless of their concrete definition.
+Dually, the type |NOFUNC| denotes the infinite intersection (i.e., subtype) of all :ref:`function types <syntax-functype>`, regardless of their concrete definition.
+This type has no values.
```
tlively, line 111: It seems complicated to describe these in terms of unions and intersections. Would dropping that part and just using supertype and subtype to describe them work?

document/core/syntax/types.rst
```diff
@@ -73,39 +73,84 @@ Conventions
...
+
+The abstract type |ANY| denotes the infinite union of all aggregate types, as well as possibly abstract values produced by *internalizing* an external reference of type |EXTERN|.
+Dually, the type |NONE| denotes the infinite intersection of all forms of aggregate types.
+This type has no values.
+
+The abstract type |EQ| is a subtype of |ANY| that includes all types for which references can be compared, i.e., aggregate values and |I31|.
+
+The abstract types |STRUCT| and |ARRAY| denote the infinite union of all :ref:`structure <syntax-structtype>` and :ref:`array <syntax-arraytype>` aggregates, respectively.
+
+The abstract type |I31| denotes *unboxed scalars*, that is, integers injected into references.
```
tlively, line 126: Is i31 really abstract? It seems that it should be considered concrete since there are values for which it is the dynamic type. Maybe instead of "concrete" we should be using "defined" or similar?

document/core/syntax/types.rst
```diff
@@ -73,39 +73,84 @@ Conventions
...
+   Engines need to perform some form of *pointer tagging* to achieve this,
+   which is why 1 bit is reserved.
 
-The type |EXTERN| denotes the infinite union of all objects owned by the :ref:`embedder <embedder>` and that can be passed into WebAssembly under this type.
 
-A *concrete* heap type consists of a :ref:`type index <syntax-typeidx>` and classifies an object of the respective :ref:`type <syntax-type>` defined in some module.
+A concrete heap type consists of a :ref:`type index <syntax-typeidx>` and classifies an object of the respective :ref:`type <syntax-type>` defined in some module.
 
-A concrete heap type can also consist of a :ref:`function type <syntax-functype>` directly.
+A concrete heap type can also consist of a :ref:`defined type <syntax-deftype>` directly.
```
tlively, line 138: What's the useful difference between saying it can consist of a type index or consist of a type directly? Why do we need to say both?

rossberg (commented):
> Is there a good way to view the rendered versions of spec PRs like this?

Unfortunately, no, AFAICT, other than building it locally. Would be great to have infrastructure for that, but I don't know how that could be built and where the rendered pages could be uploaded canonically.

document/core/syntax/types.rst
tlively: Maybe "declared" and "undeclared"?  "undeclared" seems accurate to describe all the built-in types that are not declared/defined in the type section, but idk if we have precedent talking about declarations at all.

rossberg: Right, I think we don't currently talk about "declarations" anywhere. But more importantly, "undeclared" still has this strong connotation of being erroneous, doesn't it?

tlively: Hmm maybe, although not nearly as much as "undefined" IMO. Maybe "declared" and "built in" would work?

Anyway, I don't think figuring out the perfect names here should block landing this.

tlively (approved):

PR #379, opened 2023-05-17
Update JS API

rossberg:
Adjust JS API to match core spec:

* Distinguish ref.host and ref.extern and change conversion functions accordingly.
* Add previously missing type checks to ToWebAssemblyValue.
* Remove in/externalize functions, which are not actually needed.

### DIFF
--- a/document/js-api/index.bs
+++ b/document/js-api/index.bs
@@ -80,13 +80,12 @@ urlPrefix: https://webassembly.github.io/spec/core/; spec: WebAssembly; type: df
         text: f64.const
         text: v128.const
         text: ref.null
-        text: ref.func
-        text: ref.extern
-        <!-- FIXME: correct these links once the GC proposal formal spec is complete -->
         text: ref.i31
         text: ref.array
         text: ref.struct
+        text: ref.func
         text: ref.host
+        text: ref.extern
     text: function index; url: syntax/modules.html#syntax-funcidx
     text: function instance; url: exec/runtime.html#function-instances
     text: store_init; url: appendix/embedding.html#embed-store-init
@@ -116,23 +115,15 @@ urlPrefix: https://webassembly.github.io/spec/core/; spec: WebAssembly; type: df
     text: global_read; url: appendix/embedding.html#embed-global-read
     text: global_write; url: appendix/embedding.html#embed-global-write
     text: error; url: appendix/embedding.html#embed-error
-    <!-- FIXME: these links should be updated with the GC proposal formal spec -->
-    text: i31_new; url: appendix/embedding.html#embed-gc
-    text: i31_get; url: appendix/embedding.html#embed-gc
-    text: ref_test; url: appendix/embedding.html#embed-gc
-    text: extern_internalize; url: appendix/embedding.html#embed-gc; for: embedder
-    text: extern_externalize; url: appendix/embedding.html#embed-gc; for: embedder
-    text: extern.internalize; url: syntax/instructions.html#reference-instructions
-    text: extern.externalize; url: syntax/instructions.html#reference-instructions
     text: store; url: exec/runtime.html#syntax-store
     text: table type; url: syntax/types.html#syntax-tabletype
     text: table address; url: exec/runtime.html#syntax-tableaddr
     text: function address; url: exec/runtime.html#syntax-funcaddr
     text: memory address; url: exec/runtime.html#syntax-memaddr
     text: global address; url: exec/runtime.html#syntax-globaladdr
-    text: extern address; url: exec/runtime.html#syntax-externaddr
-    <!-- FIXME: Update this link with the GC proposal formal spec -->
-    text: object address; url: exec/runtime.html#syntax-objectaddr
+    text: struct address; url: exec/runtime.html#syntax-structaddr
+    text: array address; url: exec/runtime.html#syntax-arrayaddr
+    text: host address; url: exec/runtime.html#syntax-hostaddr
     url: syntax/types.html#syntax-numtype
         text: i32
         text: i64
@@ -264,9 +255,9 @@ Each [=agent=] is associated with the following [=ordered map=]s:
     * The <dfn>Memory object cache</dfn>, mapping [=memory address=]es to {{Memory}} objects.
     * The <dfn>Table object cache</dfn>, mapping [=table address=]es to {{Table}} objects.
     * The <dfn>Exported Function cache</dfn>, mapping [=function address=]es to [=Exported Function=] objects.
-    * The <dfn>Exported GC Object cache</dfn>, mapping [=object address=]es to [=Exported GC Object=] objects.
+    * The <dfn>Exported GC Object cache</dfn>, mapping [=struct address=]es and [=array address=]es to [=Exported GC Object=] objects.
     * The <dfn>Global object cache</dfn>, mapping [=global address=]es to {{Global}} objects.
-    * The <dfn>Extern value cache</dfn>, mapping [=extern address=]es to values.
+    * The <dfn>Host value cache</dfn>, mapping [=host address=]es to values.
 
 <h2 id="webassembly-namespace">The WebAssembly Namespace</h2>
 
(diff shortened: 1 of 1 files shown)


document/js-api/index.bs
```diff
@@ -1149,27 +1143,42 @@ The algorithm <dfn>ToWebAssemblyValue</dfn>(|v|, |type|) coerces a JavaScript va
             1. Return [=ref.null=] |heaptype|.
         1. Otherwise,
             1. Throw a {{TypeError}}.
-    1. If |type| is a subtype of [=ref=] |null| [=heap-type/func=],
+    1. If |heaptype| is a subtype of [=heap-type/func=],
```
tlively, line 1146: Should we use the "matches" terminology to match the core spec?

document/js-api/index.bs
```diff
@@ -1149,27 +1143,42 @@ The algorithm <dfn>ToWebAssemblyValue</dfn>(|v|, |type|) coerces a JavaScript va
...
+            1. Let |i32| be ? [$ToInt32$](|v|).
+            1. If |v| is equal to |i32| and [=ℝ=](|v|) < 2<sup>30</sup> and [=ℝ=](|v|) ⩾ -2<sup>30</sup>,
+                1. If |heaptype| is not a supertype of [=i31=],
+                    1. Throw a {{TypeError}}.
+                1. Return [=ref.i31=] |v|.
+        1. If |v| is an [=Exported GC Object=],
+            1. Let |objectaddr| be the value of |v|'s \[[ObjectAddress]] internal slot.
+            1. Let |objecttype| be the value of |v|'s \[[ObjectType]] internal slot.
+            1. If |objecttype| is "array",
+                1. If |heaptype| is not a supertype of [=array=],
```
tlively, line 1164: Supertype or subtype, right? Since we want to do implicit downcasts on the boundary. Same for structs below.

We also need to handle the possibility of the downcast failing for structs, arrays, and functions.

document/js-api/index.bs
```diff
@@ -1149,27 +1143,42 @@ The algorithm <dfn>ToWebAssemblyValue</dfn>(|v|, |type|) coerces a JavaScript va
...
+                    1. Throw a {{TypeError}}.
+                1. Return [=ref.array=] |objectaddr|.
+            1. If |objecttype| is "struct",
+                1. If |heaptype| is not a supertype of [=struct=],
+                    1. Throw a {{TypeError}}.
+                1. Return [=ref.struct=] |objectaddr|.
+        1. If |heaptype| is not a supertype of [=any=],
```
tlively, line 1171: It would be clearer to use (in)equality here.

document/js-api/index.bs
rossberg: Done.

document/js-api/index.bs
rossberg: This is the check for the downcast. The heaptype is the target type here, so the actual type must be a subtype of it for that cast to succeed.

Using "matching" terminology now, which may make this somewhat clearer.

document/js-api/index.bs
rossberg: Turned around now using matching terminology. Not using inequality mainly for symmetry with the other checks.

document/js-api/index.bs
```diff
@@ -1149,27 +1143,42 @@ The algorithm <dfn>ToWebAssemblyValue</dfn>(|v|, |type|) coerces a JavaScript va
...
+                1. If [=i31=] does not match |heaptype|,
+                    1. Throw a {{TypeError}}.
+                1. Return [=ref.i31=] |v|.
+        1. If |v| is an [=Exported GC Object=],
+            1. Let |objectaddr| be the value of |v|'s \[[ObjectAddress]] internal slot.
+            1. Let |objecttype| be the value of |v|'s \[[ObjectType]] internal slot.
+            1. If |objecttype| is "array",
+                1. If [=array=] does not match |heaptype|,
+                    1. Throw a {{TypeError}}.
+                1. Return [=ref.array=] |objectaddr|.
```
tlively, line 1166: Don't we still need to do a ref.test to make sure the downcast succeeds when |heaptype| is a specific array type? Same with structs below.

rossberg: Well, we do not support concrete array types in the API yet, so this assumes the invariant that heaptype is one of the abstract heap types. Should we make this more powerful?

tlively: Hmm, am I misremembering the result of those discussions? I thought we allowed arbitrary types on the boundary.

tlively: Yes, I think I was correct: https://github.com/WebAssembly/gc/issues/279#issuecomment-1349574128

rossberg: Or maybe I misremember -- I assumed concrete types cannot come up because we cannot express them API-side. But did we decide to allow calling into exported Wasm functions with concrete types?

I guess there is nothing else preventing it anyway, so something needs to be fixed here. I'll extend the semantics. However, this will probably require replacing the existing [[ObjectKind]] thingy with a proper call to the embedder interface (to get the precise object type), which I haven't written yet.

tlively: Ah, right, we don't have a way to describe concrete types from the JS API, but yes, I was thinking about the case of calling into an exported function (or setting an exported global, etc) that expects a concrete type.

rossberg: Ah yes, you are right, I already added that as part of the funcref proposal.

rossberg: Hm, actually I didn't. So that probably is a dangling ref at the moment.

document/js-api/index.bs
```diff
@@ -1149,27 +1143,42 @@ The algorithm <dfn>ToWebAssemblyValue</dfn>(|v|, |type|) coerces a JavaScript va
...
+        1. If |v| [=is a Number=],
+            1. Let |i32| be ? [$ToInt32$](|v|).
+            1. If |v| is equal to |i32| and [=ℝ=](|v|) < 2<sup>30</sup> and [=ℝ=](|v|) ⩾ -2<sup>30</sup>,
+                1. If [=i31=] does not match |heaptype|,
+                    1. Throw a {{TypeError}}.
+                1. Return [=ref.i31=] |v|.
+        1. If |v| is an [=Exported GC Object=],
+            1. Let |objectaddr| be the value of |v|'s \[[ObjectAddress]] internal slot.
+            1. Let |objectkind| be the value of |v|'s \[[ObjectKind]] internal slot.
+            1. If |objecttype| is "array",
```
takikawa, line 1163: Possible typo: should this be `|objectkind|` instead of `|objecttype|`?

rossberg: Right, fixed.

document/js-api/index.bs
rossberg: Okay, I think this is fixed now. I also refactored the function to have less duplication.

document/js-api/index.bs
```diff
@@ -1138,47 +1140,35 @@ The algorithm <dfn>ToWebAssemblyValue</dfn>(|v|, |type|) coerces a JavaScript va
...
         1. [=map/Set=] |map|[|hostaddr|] to |v|.
-        1. Return [=ref.host=] |hostaddr|.
-    1. If |heaptype| matches [=extern=],
-        1. Let |r| be [=ToWebAssemblyValue=](|v|, [=ref=] |null| [=any=]).
-        1. Return [=ref.extern=] |r|.
+        1. Let |r| be [=ref.host=] |hostaddr|.
+    1. Let |store| be the current agent's [=associated store=].
+    1. Let |actualtype| be [=ref_type=](|store|, |r|).
+    1. If [=match_valtype=](|type|, [=ref=] |null| [=heap-type/func=]) is false,
+        1. Throw a {{TypeError}}.
```
tlively: Is this supposed to use `actualtype` somewhere? Also, why is the store involved?

rossberg: Oops, yes, forgot to update that line after copy & paste. Fixed.

The store is needed to obtain the precise type of a reference.

tlively (approved):
LGTM!

Issue #361, opened 2023-03-22
Text format: field names


tlively:
There's some discussion of this on #333, but it deserves its own issue.

On one hand, it would be nice for debugging and readability if the same field name could be reused across different struct types, but on the other hand this kind of dependent look up would be more complicated than how the text format currently treats names and might require further complications if we ever add unannotated accessor instructions in the future.

kripken:
Can you please explain a little more about the complexity here?

Naively, when seeing `(struct.get $Type 7` the parser needs to understand that this means field number 7 in type `$Type`, so to find the type of the field being read a lookup is needed using the type. And `(struct.get $Type $foo` seems to have the same complexity, that is, it also needs to do an operation that figures out information about the field from the combination of type + identifier. That is, a dependent look up is needed either way?

I also don't follow the point about unannotated accessor functions. Those annotations are only in the binary format, aren't they? But this discussion is about the text format.

tlively:
When field names have to be globally unique, the parser only has to bind `$foo` to an index (e.g. 7) in the index space of fields. Wherever `$foo` appears where a field index is required, the parser blindly replaces it with 7. The parser never has to know what type we are indexing into at all. Actually looking up the type definition and indexing into it isn't necessary until validation.

(This simplification does not affect the complexity of Binaryen's parsers because they collect type information as they parse rather than as a separate step.)

For the point about unannotated accessor functions, consider a hypothetical instruction `struct.get_unannotated` that in the binary format has a field index immediate but no type annotation immediate. You would expect the text format for this instruction to look like `struct.get_unannotated $foo` where `$foo` is a field name. This Just Works if `$foo` is globally unique because it can be parsed as `struct.get_unannotated 7`, but if different types can contain `$foo` fields at different indices, then either we force parsers to track type information so it knows which `$foo` this is or we complicate the text format by inserting extra type information that isn't present in the binary.

kripken:
I see, thanks. For comparison, I believe all production binary parsers parse+validate at the same time, so if text parsers did the same then neither of the problems you mentioned would be an issue, I think? Is it unreasonable for some reason to expect them to do so?

My perspective on this is that we toolchain developers spend a lot of time debugging using wat files. So any work that goes into making easier debugging there will pay dividends over time for anyone writing a compiler or toolchain for wasm, and that number of people is larger than the number that write wat parsers I think. But I may be missing something.

tlively:
[...]

PR #385, opened 2023-06-12
Spec text format

rossberg:
Also, to sync up, implement dependent name spaces for fields in interpreter.


document/core/text/modules.rst
```diff
@@ -96,18 +83,23 @@ If inline declarations are given, then their types must match the referenced :re
...
-        I.\ITYPEDEFS[x] = [t_1^\ast] \toF [t_2^\ast] \wedge
+        I.\ITYPEDEFS[x] = \TSUB~\TFINAL~(\TFUNC~[t_1^\ast] \toF [t_2^\ast]) \wedge
         I' = \{\ILOCALS~\F{id}(\Tparam)^\ast\} \idcwellformed) \\
         \end{array} \\
    \end{array}
 
+.. note::
+   If inline declarations are given, their types must be *syntactically* equal to the types from the indexed definition;
+   possible type :ref:`substitutions <notation-subst>` from other definitions that might make them equal are not taken into account.
+   This is to simplify syntactic pre-processing in the presence of recursive types.
```
tlively, line 101: Can you give an example of what this allows and disallows? I don't grok the implications yet.

document/core/text/modules.rst
```diff
@@ -138,11 +130,11 @@ In that case, a :ref:`type index <syntax-typeidx>` is automatically inserted:
      \text{(}~\text{type}~~x~\text{)}~~\Tparam^\ast~~\Tresult^\ast \\
    \end{array}
 
-where :math:`x` is the smallest existing :ref:`type index <syntax-typeidx>` whose definition in the current module is the :ref:`function type <syntax-functype>` :math:`[t_1^\ast] \toF [t_2^\ast]`.
-If no such index exists, then a new :ref:`type definition <text-type>` of the form
+where :math:`x` is the smallest existing :ref:`type index <syntax-typeidx>` whose definition in the current module is the :ref:`final <syntax-final>` :ref:`function type <syntax-functype>` :math:`[t_1^\ast] \toF [t_2^\ast]`.
+If no such index exists, then a new :ref:`recursive type <text-rectype>` of the form
```
tlively, line 134: Does this capture the constraint that only types in singleton rec groups are eligible to be found? IIRC, that's how we decided this should work.

test/core/gc/struct.wast
```diff
@@ -50,6 +43,28 @@
...
+;; Field names
+
+(module
+  (type (struct (field $x i32)))
+  (type $t1 (struct (field i32) (field $x f32)))
+  (type $t2 (struct (field i32 i32) (field $x i64)))
+
+  (func (param (ref 0)) (result i32) (struct.get 0 $x (local.get 0)))
+  (func (param (ref $t1)) (result f32) (struct.get 1 $x (local.get 0)))
+  (func (param (ref $t2)) (result i64) (struct.get $t2 $x (local.get 0)))
```
tlively, line 55: It would be good to have tests for all combinations of #names and raw numbers for the type index and the field index if we don't have that already. Did you decide to do anything to anticipate having instructions with field indices but no type indices?

tlively (commented):
Does this contain the abbreviation allowing fields to be combined, e.g. `(field i32 i64)`? I didn't see that when reading through this PR.

document/core/text/modules.rst
rossberg: For example, if you have
```
(type $s1 (struct))
(type $s2 (struct))
(type $t (func (param (ref $s1))))
(func $f (param (ref $s2)))
```
then it would not pick up $t but introduce a new $t that is using $t2. This is so that desugaring does not depend on type canonicalisation. This choice doesn't affect typability, and the difference is not observable for well-formed modules, unless you were to inspect the desugared syntax itself to discover the additional type entries.

document/core/text/modules.rst
rossberg: Ah, you are right, that's missing. Fixed and made more precise.

test/core/gc/struct.wast
rossberg: I added a simple test for each combination. (To be honest, we could probably use more tests for the various struct/array instructions in general.)

In terms of anticipation, no, what do you have in mind?

rossberg (commented):
Comments addressed, PTAL.

document/core/text/modules.rst
tlively: Makes sense, thanks!

test/core/gc/struct.wast
tlively: I had one idea for future-proofing the syntax described in the second paragraph here: https://github.com/WebAssembly/gc/issues/361#issuecomment-1481597070

tlively (approved):
LGTM % my previous comment about the `(field i32 i64 ...)` abbreviation.

test/core/gc/struct.wast
rossberg: Can you elaborate on that idea? I'm not really understanding why the syntax `struct.get t (field t i)` is more future-proof than `struct.get t i`. What does repeating the type twice buy us?

rossberg:
> LGTM % my previous comment about the `(field i32 i64 ...)` abbreviation.

Oops, sorry, had missed that comment. Indeed, added now.

test/core/gc/struct.wast
tlively: In the future if we have some `struct.get_unannotated <fieldidx>` instruction that has no type annotation, and if the text syntax for `fieldidx` contains a type index along with the symbolic field name, then we can still unambiguously figure out how to resolve the symbolic field name to an index. For the current instructions, which already contain type indices, the extra type index in the fieldidx syntax is redundant, so we can define abbreviations to make the repetition unnecessary.

rossberg: Why couldn't we just use the text syntax `struct.get_unannotated tidx? $x` in that case? (Allowing omitting the type index only when the field index is numeric.)

tlively: That would work, too! If you'd be happy with that, then it sounds good to me.

PR #391, opened 2023-06-27
Fix spec.fix-array.new_data

rossberg:
See https://github.com/WebAssembly/gc/pull/387#discussion_r1242864600


tlively (approved):
Makes sense 👍

PR #402, opened 2023-07-19
Fix mixed use of ft^\ast and ft^n in struct.get/set execution

### LINKED ISSUE #400: Inconsistent use of \X{ft}^n and \X{ft}^\ast in execution spec?
For example, the execution semantics for `struct.new` and `struct.new_default` use `ft^*` in the prose and `ft^n` in the formalism, whereas `struct.get` and `struct.set` use `ft^*` throughout.

Is the rule that if the length matters, use `ft^n` and otherwise use `ft^*`? If so, is it intentional that the prose uses `ft^*` in either case?

tlively:
Consistently use `\X{ft}^n` to refer to the struct field types in the execution
rules for struct.get and struct.set.

Fixes #400.

### DIFF
--- a/document/core/exec/instructions.rst
+++ b/document/core/exec/instructions.rst
@@ -562,7 +562,7 @@ Reference Instructions
      &
      \begin{array}[t]{@{}r@{~}l@{}}
       (\iff & \expand(F.\AMODULE.\MITYPES[x]) = \TSTRUCT~\X{ft}^n \\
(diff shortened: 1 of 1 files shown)


tlively:
* main
  * **PR #402** <a href="https://app.graphite.dev/github/pr/WebAssembly/gc/402" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/WebAssembly/gc/402?utm_source=stack-comment).

rossberg (approved):
Thanks!

rossberg (approved):
Thanks!

PR #406, opened 2023-07-20
Extend soundness appendix

rossberg:
Also, kill a few related todos.

@conrad-watt, PTAL.


document/core/appendix/properties.rst
```diff
@@ -382,7 +658,7 @@ Module instances are classified by *module contexts*, which are regular :ref:`co
 :ref:`Module Instances <syntax-moduleinst>` :math:`\moduleinst`
 ...............................................................
 
-* Each :ref:`defined type <syntax-deftype>` :math:`\deftype_i` in :math:`\moduleinst.\MITYPES` must be :ref:`valid <valid-deftype>`.
+* Each :ref:`defined type <syntax-deftype>` :math:`\deftype_i` in :math:`\moduleinst.\MITYPES` must be :ref:`valid <valid-deftype>` under the empty :ref:`context <context>`.
```
conrad-watt, line 661: To confirm, types in module instances aren't stored in their rolled form?

document/core/appendix/properties.rst
```diff
@@ -128,23 +313,26 @@ Module instances are classified by *module contexts*, which are regular :ref:`co
...
+       \SSTRUCTS~\structinst^\ast,
+       \SARRAYS~\arrayinst^\ast \}
        \end{array}
      \end{array}
    }{
      \vdashstore S \ok
    }
 
-.. note::
-   The validity condition on type instances ensures the absence of cyclic types.
```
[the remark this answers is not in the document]
rossberg: This note has been outdated for a while, I just missed it. We no longer have type instances. And since we no longer have indirections through a type store, types are closed syntactic terms and hence non-cyclic by construction.

document/core/appendix/properties.rst
rossberg: No, they are assumed to be rolled up (and closed), see also the definition of alloctype (which I tried to make a bit more readable). But of course, in terms of wf rules for instances, that does not materialise as an observable constraint, since any 1-step unrolling is also a valid defined type, and you can construct a(nother) module that produces that.

For example,
```
(module $A
  (rec (type $t1 (array (ref $t1)))
  (rec (type $t2 (array (ref $t2)))
)

(module $B
  (rec (type $t1 (array (ref $t1)))
  (rec (type $t2' (array (ref $t1)))
)
```
are both valid modules, but in the instance of $B, the deftype for $t2' would to look exactly like the unrolling of $t2 in $A.

rossberg:
@conrad-watt, I also added the cycle-freedom condition.

conrad-watt (approved):
lgtm (although I didn't previously spot the tail call omissions)

PR #411, opened 2023-07-24
Fix JS integer conversion

### LINKED ISSUE #399: Clarification on ToWasmValue for i31ref
[Source](https://webassembly.github.io/gc/js-api/index.html#towebassemblyvalue).

```
Else if v [is a Number](https://tc39.github.io/ecma262/#sec-ecmascript-language-types-number-type) and v is equal to ? [ToInt32](https://tc39.es/ecma262/multipage/abstract-operations.html#sec-toint32)(v) and [ℝ](https://tc39.github.io/ecma262/#%E2%84%9D)(v) < 230 and [ℝ](https://tc39.github.io/ecma262/#%E2%84%9D)(v) ⩾ -230,

    Let r be [ref.i31](https://webassembly.github.io/spec/core/exec/runtime.html#values) v.
```

For the final step, we construct a `ref.i31`, which takes a `u31` parameter, with `v` which is a JS number. From the pre-conditions, `v` is an integer within the range [-2^30, 2^30), so it may be negative. What happens to negative Numbers when they're converted to u31? It feels like an explicit conversion is missing here from JS number to u31.

rossberg:
Fix #399 and a couple of other small bugs with conversion to/from JS numbers.

### DIFF
--- a/document/js-api/index.bs
+++ b/document/js-api/index.bs
@@ -1079,10 +1079,12 @@ Note: Exported Functions do not have a \[[Construct]] method and thus it is not 
 The algorithm <dfn>ToJSValue</dfn>(|w|) coerces a [=WebAssembly value=] to a JavaScript value by performing the following steps:
 
 1. Assert: |w| is not of  the form [=v128.const=] <var ignore>v128</var>.
-1. If |w| is of the form [=i64.const=] |i64|,
-    1. Let |v| be [=signed_64=](|i64|).
-    1. Return [=ℤ=](|v| interpreted as a mathematical value).
-1. If |w| is of the form [=i32.const=] |i32|, return [=𝔽=]([=signed_32=](|i32| interpreted as a mathematical value)).
+1. If |w| is of the form [=i64.const=] |u64|,
+    1. Let |i64| be [=signed_64=](|u64|).
+    1. Return [=ℤ=](|i64| interpreted as a mathematical value).
+1. If |w| is of the form [=i32.const=] |u32|,
+    1. Let |i32| be [=signed_32=](|i32|).
+    2. Return [=𝔽=](|i32| interpreted as a mathematical value).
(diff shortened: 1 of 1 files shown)


document/js-api/index.bs
```diff
@@ -1145,8 +1151,10 @@ The algorithm <dfn>ToWebAssemblyValue</dfn>(|v|, |type|) coerces a JavaScript va
         1. Return [=ref.extern=] |r|.
     1. If |v| is null,
         1. Let |r| be [=ref.null=] |heaptype|.
-    1. Else if |v| [=is a Number=] and |v| is equal to ? [$ToInt32$](|v|) and [=ℝ=](|v|) < 2<sup>30</sup> and [=ℝ=](|v|) ⩾ -2<sup>30</sup>,
-        1. Let |r| be [=ref.i31=] |v|.
+    1. Else if |v| [=is a Number=] and |v| is equal to [=?=] [$ToInt32$](|v|) and [=ℝ=](|v|) < 2<sup>30</sup> and [=ℝ=](|v|) ⩾ -2<sup>30</sup>,
+        1. Let |i31| [=?=] [$ToInt32$](|v|).
+        1. Let |u31| be the unsigned integer such that |i31| is [=signed_32=](|i|).
```
eqrion, line 1156: Did you mean, `[=signed_32=](|i31|).` here?

eqrion: Hmm actually, is `signed_32` right here?

For example, with `i31 = -1` we must therefore have `u31 = (2^32) - 1` which is larger than the highest u31 value.

rossberg: Yes, you are right, has to be signed_31. Fixed.

tlively:
I don't have enough expertise with the JS API to review a change in the details like this, so LGTM if it looks good to @eqrion.

Issue #333, opened 2022-11-01
Text format: type definition abbreviations


tlively:
Since we write a lot of text format tests in Binaryen, we've implemented a number of abbreviations to make type definitions shorter. Here's the grammar I've implemented the new Binaryen parser that combines the text format used in this repo with the abbreviations used in Binaryen.

```
deftype ::= '(' 'rec' subtype* ')'
          | subtype

subtype ::= '(' 'type' id? '(' 'sub' typeidx? strtype ')' ')'
          | '(' 'type' id? strtype ')'

strtype ::= functype
          | structtype
          | arraytype

functype ::= '... ;; same as MVP

structtype ::= '(' 'struct' field* ')'

arraytype ::= '(' 'array' field ')'

field ::= '(' 'field' id? fieldtype ')'
        | '(' 'field' fieldtype* ')'

fieldtype ::= storagetype
            | '(' 'mut' storagetype ')'

storagetype ::= valtype | packedtype

packedtype ::= 'i8' | 'i16'
```

Comments and suggested edits welcome. Once it looks good, we should document it in the MVP doc. For the actual spec, we'll have to extract the abbreviations from the grammar and turn them into rewrite rules.

sbc100:
What is `strtype`?  Presumably nothing to do with strings, but it certainly reads like it is.  Maybe a better name?

tlively:
`strtype` is short for "structural type" and comes directly from the proposal docs: https://github.com/WebAssembly/gc/blob/main/proposals/gc/MVP.md#type-definitions. I agree it would be nice to find a clearer name or abbreviation for those, but I do want to keep the explanation of the text format in sync with the abstract syntax.

rossberg:
Actually, strtype stands for "structured type", i.e., a compound type like structs, funcs, arrays, as opposed to scalar types like ints and floats.

tlively:
There is also the issue of using symbolic references to struct fields, as discussed here: https://github.com/WebAssembly/gc/issues/193#issuecomment-843783920. I don't believe Binaryen's parser currently allows symbolic field references, but I agree they would be nice to have. Since we ended up keeping the type annotations on `struct.get` and friends, we could keep the dependent name lookup behavior where each struct type gets its own index space of types. @rossberg, wdyt? What if separate definitions of the same type contain conflicting field names?

rossberg:
> Binaryen supports this today (although it doesn’t allow using field names symbolically in the text)

I'm confused, where does it support them then?

[...]

PR #413, opened 2023-07-28
Use `open` instead of `final` in the text format

bvisness:
Addresses discussion from #333. With this patch, the keyword `final` in subtypes has been replaced with the keyword `open`:

```
(type $s1 (sub (struct)))                       ;; final with no super types
(type $s2 (sub open (struct)))                  ;; non-final with no super types
(type $s3 (sub $s2 (struct (field i32))))       ;; final with one super type
(type $s4 (sub open $s2 (struct (field i32))))  ;; non-final with one super type

(type $s5 (struct))                             ;; equivalent to $s1 (final)
```

This avoids a confusing flip-flopping behavior where a seemingly-unnecessary empty `sub` was commonly used to make types non-final. An empty `sub` is now equivalent to no `sub`, making the abbreviation more obvious.

This patch is currently quite minimal, and only modifies enough of the spec and interpreter to support `open` in the text format. The word `final` is otherwise used throughout the spec and interpreter. If we would rather use the word `open` everywhere, I'd be happy to update the rest of the spec and interpreter to match.

### DIFF
--- a/document/core/text/types.rst
+++ b/document/core/text/types.rst
@@ -85,8 +85,8 @@ Abbreviations
 There are shorthands for references to abstract heap types.
 
 .. math::
-   \begin{array}{llclll}
-   \production{reference type} &
+   \begin{array}{lcl}
+   \production{reference type}
      \text{anyref} &\equiv& \text{(}~\text{ref}~~\text{null}~~\text{any}~\text{)} \\
      \text{eqref} &\equiv& \text{(}~\text{ref}~~\text{null}~~\text{eq}~\text{)} \\
      \text{i31ref} &\equiv& \text{(}~\text{ref}~~\text{null}~~\text{i31}~\text{)} \\
@@ -255,30 +255,33 @@ Recursive Types
      \text{(}~\text{type}~~\Tid^?~~\X{st}{:}\Tsubtype_I~\text{)}
        &\Rightarrow& \X{st} \\
    \production{sub type} & \Tsubtype_I &::=&
-     \text{(}~\text{sub}~~\text{final}^?~~x^\ast{:\,}\Tvec(\Ttypeidx_I)~~\X{ct}{:}\Tcomptype_I~\text{)}
+     \text{(}~\text{sub}~~\text{open}^?~~x^\ast{:\,}\Tvec(\Ttypeidx_I)~~\X{ct}{:}\Tcomptype_I~\text{)}
        &\Rightarrow& \TSUB~\TFINAL^?~x^\ast~\X{ct} \\
    \end{array}
+
+.. note::
+   A subtype is :ref:`final <syntax-subtype>` if the :math:`\text{open}` keyword is not present, and vice versa.
 
 
 Abbreviations
 .............
 
-Singular recursive types can omit the |Trec| keyword:
+Singular recursive types can omit the :math:`\text{rec}` keyword:
 
 .. math::
-   \begin{array}{llclll}
+   \begin{array}{lcl}
    \production{recursive type} &
-     \Tsubtype &\equiv&
-     \text{(}~~\text{rec}~~\Tsubtype~~\text{)} \\
+     \Tdeftype &\equiv&
+     \text{(}~~\text{rec}~~\Tdeftype~~\text{)} \\
    \end{array}
 
-Similarly, final sub types with no super-types can omit the |Tsub| keyword and arguments:
+Similarly, final sub types with no super-types can omit the :math:`\text{sub}` keyword:
 
 .. math::
-   \begin{array}{llclll}
+   \begin{array}{lcl}
    \production{sub type} &
      \Tcomptype &\equiv&
-     \text{(}~~\text{sub}~~\text{final}~~\epsilon~~\Tcomptype~~\text{)} \\
+     \text{(}~~\text{sub}~~\Tcomptype~~\text{)} \\
    \end{array}
 
 
--- a/interpreter/syntax/types.ml
+++ b/interpreter/syntax/types.ml
@@ -308,8 +308,8 @@ let string_of_null = function
   | Null -> "null "
 
 let string_of_final = function
-  | NoFinal -> ""
-  | Final -> " final"
+  | NoFinal -> " open"
+  | Final -> ""
 
 let string_of_mut s = function
   | Cons -> s
--- a/interpreter/text/arrange.ml
+++ b/interpreter/text/arrange.ml
@@ -78,8 +78,8 @@ let val_type t = string_of_val_type t
 let storage_type t = string_of_storage_type t
 
 let final = function
-  | NoFinal -> ""
-  | Final -> " final"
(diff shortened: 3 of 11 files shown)


document/core/text/types.rst
```diff
@@ -255,30 +255,33 @@ Recursive Types
...
+Singular recursive types can omit the :math:`\text{rec}` keyword:
 
 .. math::
-   \begin{array}{llclll}
+   \begin{array}{lcl}
    \production{recursive type} &
-     \Tsubtype &\equiv&
-     \text{(}~~\text{rec}~~\Tsubtype~~\text{)} \\
+     \Tdeftype &\equiv&
+     \text{(}~~\text{rec}~~\Tdeftype~~\text{)} \\
```
bvisness, line 275: This is not related to my work, but I believe it to be a bug in the spec that is easy to fix.

document/core/text/types.rst
```diff
@@ -85,8 +85,8 @@ Abbreviations
 There are shorthands for references to abstract heap types.
 
 .. math::
-   \begin{array}{llclll}
-   \production{reference type} &
+   \begin{array}{lcl}
+   \production{reference type}
```
bvisness, line 89: This is also unrelated to my work, but this table was jank, and now it is not.

document/core/text/types.rst
```diff
@@ -255,30 +255,33 @@ Recursive Types
...
-Similarly, final sub types with no super-types can omit the |Tsub| keyword and arguments:
+Similarly, final sub types with no super-types can omit the :math:`\text{sub}` keyword:
 
 .. math::
-   \begin{array}{llclll}
+   \begin{array}{lcl}
    \production{sub type} &
      \Tcomptype &\equiv&
-     \text{(}~~\text{sub}~~\text{final}~~\epsilon~~\Tcomptype~~\text{)} \\
+     \text{(}~~\text{sub}~~\Tcomptype~~\text{)} \\
```
bvisness, line 284: It's not clear to me why the `epsilon` was in here, but I can add it back if it was actually necessary.

tlively: It probably represented the empty vector of supertypes. I think it would make sense to put it and the "and arguments" language back.

document/core/text/types.rst
```diff
@@ -255,30 +255,33 @@ Recursive Types
      \text{(}~\text{type}~~\Tid^?~~\X{st}{:}\Tsubtype_I~\text{)}
        &\Rightarrow& \X{st} \\
    \production{sub type} & \Tsubtype_I &::=&
-     \text{(}~\text{sub}~~\text{final}^?~~x^\ast{:\,}\Tvec(\Ttypeidx_I)~~\X{ct}{:}\Tcomptype_I~\text{)}
+     \text{(}~\text{sub}~~\text{open}^?~~x^\ast{:\,}\Tvec(\Ttypeidx_I)~~\X{ct}{:}\Tcomptype_I~\text{)}
        &\Rightarrow& \TSUB~\TFINAL^?~x^\ast~\X{ct} \\
    \end{array}
 
+.. note::
+   A subtype is :ref:`final <syntax-subtype>` if the :math:`\text{open}` keyword is not present, and vice versa.
```
rossberg, line 263: See above, this comment shouldn't be needed if the text format is consistent with the AST.

document/core/text/types.rst
```diff
@@ -255,30 +255,33 @@ Recursive Types
...
+
 
 Abbreviations
 .............
 
-Singular recursive types can omit the |Trec| keyword:
+Singular recursive types can omit the :math:`\text{rec}` keyword:
```
rossberg, line 269: Why this change?

document/core/text/types.rst
rossberg: No, I think this was correct before.

document/core/text/types.rst
```diff
@@ -255,30 +255,33 @@ Recursive Types
...
+     \Tdeftype &\equiv&
+     \text{(}~~\text{rec}~~\Tdeftype~~\text{)} \\
    \end{array}
 
-Similarly, final sub types with no super-types can omit the |Tsub| keyword and arguments:
+Similarly, final sub types with no super-types can omit the :math:`\text{sub}` keyword:
 
 .. math::
-   \begin{array}{llclll}
+   \begin{array}{lcl}
```
rossberg, line 281: I think this needs to be at least `llcl`, otherwise Latex will barf.

document/core/text/types.rst
rossberg: What @tlively said, this was intentional.

tlively:
I would support having both `sub open` and `sub final`. That seems strictly clearer, although slightly more verbose. It's also useful to still have the "final" language for non-open types.

@bvisness, will you have time to work on this more today and Monday? If not, I can help out.

tlively:
@rossberg, would it suffice to add `sub open` just to the text format and keep `final?` in the abstract syntax, or would you want to change the abstract syntax to have `final | open` as well?

rossberg:
I think we should keep them consistent. It actually simplifies some formulations as well when we have a named phrase like `ext ::= FINAL | OPEN`, because then `ext` can be used as a metavariable instead of clumsy `FINAL?`.

tlively:
@rossberg, would you be able to work on completing this change? If not, I will take a stab at it in the morning pacific time.

rossberg:
@tlively, I can try, but I can't promise anything — I'm on an a flight Odyssee right now after I was dumped from my flight back to Europe yesterday. Hence my bigger concern is that I even make it back in time for the meeting tomorrow. ;)

bvisness:
Apologies for my absence. I can resume work on this tomorrow if it's not already done.

bvisness (commented):
@tlively's work in #423 looks good to me, so this PR can probably be closed, but I did respond to a couple of the questions up above. In particular, I still feel that the one abbreviation does contain a bug.

document/core/text/types.rst
tlively: @bvisness's fix here looks right to me. We don't want to allow `(sub (struct))`, a subtype, to appear on its own, but we do want to allow `(type (sub (struct)))`, a deftype, to appear on its own.

document/core/text/types.rst
tlively: It may be useful to note that macros.def does not define `Trec`, so I think it is parsing as `|TREC|`, which does link to the abstract syntax rather than the textual keyword.

rossberg: Ah, good points.

document/core/text/types.rst
rossberg: Ah, you are right, I was confused by the impedance mismatch with the abstract syntax, where deftype is something else. We should probably not be naming this one in the text format deftype, but rather something else (typedef?).

document/core/text/types.rst
tlively: Done as a standalone fix in #426.

PR #415, opened 2023-08-14
Extends Name Section w/ Type & Field

### LINKED ISSUE #193: Extending the Name section for struct fields
For an improved debugging experience, it would be desirable to have debuggers show names for struct fields, so that developers can map objects fields they see in a Wasm debugger back to the class/etc member fields they defined in their source language. Extending the custom [name section](https://webassembly.github.io/spec/core/bikeshed/index.html#name-section%E2%91%A0) for this purpose seems straightforward: based on [the extended-name-section proposal](https://github.com/WebAssembly/extended-name-section/blob/master/document/core/appendix/custom.rst), I assume we will add:

`id = 10, fieldnamesubsec ::= namesubsection_10 (indirectnamemap)`

i.e. same structure as the "local names" subsection, just with different ID and meaning. Two-level index: type_index + field_index.

ashleynh:
This PR updates the spec to add additional subsections for type and field within the custom name section. struct.get and struct.set also updated to indicate that their field indices are of type fieldidx, instead of u32. As a result, the metavariable representation was updated from i (integers) to y (indices).

Closes #193


document/core/appendix/custom.rst
```diff
@@ -143,3 +145,33 @@ It consists of an :ref:`indirect name map <binary-indirectnamemap>` assigning lo
    \production{local name subsection} & \Blocalnamesubsec &::=&
      \Bnamesubsection_2(\Bindirectnamemap) \\
    \end{array}
+
```
rossberg, line 148: Nit:
```suggestion


```

document/core/appendix/custom.rst
```diff
@@ -143,3 +145,33 @@ It consists of an :ref:`indirect name map <binary-indirectnamemap>` assigning lo
...
+The *type name subsection* has the id 4.
+It consists of a :ref:`name map <binary-namemap>` assigning type names to :ref:`type indices <syntax-typeidx>`.
+
+.. math::
+   \begin{array}{llclll}
+   \production{type name subsection} & \Btypenamesubsec &::=&
+     \Bnamesubsection_1(\Bnamemap) \\
+   \end{array}
+
```
rossberg, line 163: Nit:
```suggestion


```

document/core/syntax/modules.rst
```diff
@@ -87,6 +91,7 @@ The index space for :ref:`locals <syntax-local>` is only accessible inside a :re
 
 Label indices reference :ref:`structured control instructions <syntax-instr-control>` inside an instruction sequence.
 
+Each :ref:`aggregate type <syntax-aggrtype>` provides an index space for its :ref:`fields <syntax-fieldtype>`.
```
rossberg, line 95: Nit:
```suggestion


```

rossberg (approved):
Note that field offsets are different in nature from other indices: in programming language terms they are numeric "labels" (in the record sense), whereas all other indices are numeric "variables". The difference is somewhat technical, though (has to do with how binding works and what can be renamed/shifted how), so I can live with calling them fieldidx.

ashleynh:
Thanks for taking a look @rossberg! Can you expand on the technical differences w.r.t. binding and renames/shifts? Would calling it fieldOffsetIdx be preferred? We could add a note calling out the how they are numeric labels in the record sense.

rossberg:
No worries, I think the name is fine. The technicalities aren't easily explained in a couple of sentences, but roughly, for variables, use site and definition site are connected by scoping, i.e., you'll look up a variable in the current scope chain. Labels on the other hand are tied to, and looked up in, the _type_ of some operand. Multiple labels of the same name can be used in the same scope while referring to different definitions (or sometimes, do not even need a definition). When you get down to something like lambda calculus, one consequence is that variables are freely renameable (and substitutable), while labels are not. But I'm digressing, not sure this is helpful... :)

ashleynh:
Very helpful explanation, thank you!

PR #416, opened 2023-08-24
Update array.copy semantics to handle packed types

tlively:
Previously the array.copy semantics were defined in terms of executing array.get
on the source array, which is not correct if the source array contains a packed
type. Update the semantics prose and formalism to choose between array.get and
array.get_u depending on whether the array contents are packed. The interpreter
does not need updating because it already handled the difference correctly.

### DIFF
--- a/document/core/exec/instructions.rst
+++ b/document/core/exec/instructions.rst
@@ -1130,11 +1130,11 @@ Reference Instructions
 
 5. Assert: due to :ref:`validation <valid-array.copy>`, a :ref:`value <syntax-val>` of :ref:`type <syntax-valtype>` :math:`\I32` is on the top of the stack.
 
-6. Pop the value :math:`n` from the stack.
+6. Pop the value :math:`\I32.\CONST~n` from the stack.
 
 7. Assert: due to :ref:`validation <valid-array.copy>`, a :ref:`value <syntax-val>` of :ref:`type <syntax-valtype>` :math:`\I32` is on the top of the stack.
 
-8. Pop the value :math:`s` from the stack.
+8. Pop the value :math:`\I32.\CONST~s` from the stack.
 
 9. Assert: due to :ref:`validation <valid-array.copy>`, a :ref:`value <syntax-val>` of :ref:`type <syntax-valtype>` :math:`(\REF~\NULL~y)` is on the top of the stack.
 
@@ -1142,7 +1142,7 @@ Reference Instructions
 
 11. Assert: due to :ref:`validation <valid-array.copy>`, a :ref:`value <syntax-val>` of :ref:`type <syntax-valtype>` :math:`\I32` is on the top of the stack.
 
-12. Pop the value :math:`d` from the stack.
+12. Pop the value :math:`\I32.\CONST~d` from the stack.
 
 13. Assert: due to :ref:`validation <valid-array.copy>`, a :ref:`value <syntax-val>` of :ref:`type <syntax-valtype>` :math:`(\REF~\NULL~x)` is on the top of the stack.
 
--- a/document/core/exec/instructions.rst
+++ b/document/core/exec/instructions.rst
@@ -1120,61 +1120,75 @@ Reference Instructions
 :math:`\ARRAYCOPY~x~y`
 ......................
 
-.. todo:: Handle packed fields correctly via array.get_u instead of array.get
+1. Assert: due to :ref:`validation <valid-array.init_data>`, the :ref:`defined type <syntax-deftype>` :math:`F.\AMODULE.\MITYPES[y]` exists.
 
-1. Assert: due to :ref:`validation <valid-array.copy>`, a :ref:`value <syntax-val>` of :ref:`type <syntax-valtype>` :math:`\I32` is on the top of the stack.
+2. Let :math:`\deftype` be the :ref:`defined type <syntax-deftype>` :math:`F.\AMODULE.\MITYPES[y]`.
 
-2. Pop the value :math:`n` from the stack.
+3. Assert: due to :ref:`validation <valid-array.init_data>`, the :ref:`expansion <aux-expand-deftype>` of :math:`\deftype` is an :ref:`array type <syntax-arraytype>`.
 
-3. Assert: due to :ref:`validation <valid-array.copy>`, a :ref:`value <syntax-val>` of :ref:`type <syntax-valtype>` :math:`\I32` is on the top of the stack.
+4. Let :math:`\TARRAY~\X{ft}` be the :ref:`expanded <aux-expand-deftype>` :ref:`array type <syntax-arraytype>` :math:`\deftype`.
 
-4. Pop the value :math:`s` from the stack.
+5. If :math:`\X{ft} = \mut~\packedtype`, then:
 
-5. Assert: due to :ref:`validation <valid-array.copy>`, a :ref:`value <syntax-val>` of :ref:`type <syntax-valtype>` :math:`(\REF~\NULL~y)` is on the top of the stack.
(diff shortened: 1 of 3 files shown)


tlively:
* main
  * **PR #416** <a href="https://app.graphite.dev/github/pr/WebAssembly/gc/416" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/WebAssembly/gc/416?utm_source=stack-comment).

document/core/exec/instructions.rst
```diff
@@ -1248,24 +1262,32 @@ Reference Instructions
...
+       (\REFARRAYADDR~a_2)~(\I32.\CONST~s)~\X{get} \\
        (\ARRAYSET~x) \\
        (\REFARRAYADDR~a_1)~(\I32.\CONST~d+1)~(\REFARRAYADDR~a_2)~(\I32.\CONST~s+1)~(\I32.\CONST~n)~(\ARRAYCOPY~x~y) \\
        \end{array}
      \\ \qquad
-     (\otherwise, \iff d \leq s)
+     \begin{array}[t]{@{}r@{~}l@{}}
+     (\otherwise, \iff & d \leq s \\
+                  \land & (\X{get} = \ARRAYGET~y \vee \X{get} = \ARRAYGETU~y) \\
+                  \land & (F.\AMODULE.\MITYPES[x] = \TARRAY~\mut~\packedtype \Longleftrightarrow \X{get} = \ARRAYGETU~y))
```
tlively, line 1273: Can you think of a better way to express this? Is there a better symbol to use for bidirectional implication than `\Longleftrightarrow`?

document/core/exec/instructions.rst
```diff
@@ -1120,61 +1120,75 @@ Reference Instructions
 :math:`\ARRAYCOPY~x~y`
 ......................
 
-.. todo:: Handle packed fields correctly via array.get_u instead of array.get
+1. Assert: due to :ref:`validation <valid-array.init_data>`, the :ref:`defined type <syntax-deftype>` :math:`F.\AMODULE.\MITYPES[y]` exists.
```
rossberg, line 1123:
```suggestion
1. Assert: due to :ref:`validation <valid-array.copy>`, the :ref:`defined type <syntax-deftype>` :math:`F.\AMODULE.\MITYPES[y]` exists.
```

document/core/exec/instructions.rst
```diff
@@ -1120,61 +1120,75 @@ Reference Instructions
...
 
-1. Assert: due to :ref:`validation <valid-array.copy>`, a :ref:`value <syntax-val>` of :ref:`type <syntax-valtype>` :math:`\I32` is on the top of the stack.
+2. Let :math:`\deftype` be the :ref:`defined type <syntax-deftype>` :math:`F.\AMODULE.\MITYPES[y]`.
 
-2. Pop the value :math:`n` from the stack.
+3. Assert: due to :ref:`validation <valid-array.init_data>`, the :ref:`expansion <aux-expand-deftype>` of :math:`\deftype` is an :ref:`array type <syntax-arraytype>`.
```
rossberg, line 1127:
```suggestion
3. Assert: due to :ref:`validation <valid-array.copy>`, the :ref:`expansion <aux-expand-deftype>` of :math:`\deftype` is an :ref:`array type <syntax-arraytype>`.
```

document/core/exec/instructions.rst
rossberg: The symbol would be correct. But I would suggest to factor out the choice of `get` into an auxiliary function which does a case distinction on the storage type, i.e., something like
```
getfield(valtype) = array.get
getfield(packedtype) = array.get_u
```

tlively: Thanks, I've written this up now. PTAL!

document/core/exec/instructions.rst
```diff
@@ -1120,75 +1120,67 @@ Reference Instructions
...
+3. Assert: due to :ref:`validation <valid-array.copy>`, the :ref:`expansion <aux-expand-deftype>` of :math:`\deftype` is an :ref:`array type <syntax-arraytype>`.
 
-   a. Let :math:`\X{get}` be the instruction :math:`\ARRAYGETU~y`.
+4. Let :math:`\TARRAY~\mut~\X{st}` be the :ref:`expanded <aux-expand-deftype>` :ref:`array type <syntax-arraytype>` :math:`\deftype`.
 
-6. Else:
+5. Assert: due to :ref:`validation <valid-array.copy>`, a :ref:`value <syntax-val>` of :ref:`type <syntax-valtype>` :math:`\I32` is on the top of the stack.
 
-   a. Let :math:`\X{get}` be the instruction :math:`\ARRAYGET~y`.
+6. Pop the value :math:`n` from the stack.
```
rossberg, line 1133:
```suggestion
6. Pop the value :math:`\I32.\CONST~n` from the stack.
```

document/core/exec/instructions.rst
```diff
@@ -1120,75 +1120,67 @@ Reference Instructions
...
 
 7. Assert: due to :ref:`validation <valid-array.copy>`, a :ref:`value <syntax-val>` of :ref:`type <syntax-valtype>` :math:`\I32` is on the top of the stack.
 
-8. Pop the value :math:`n` from the stack.
-
-9. Assert: due to :ref:`validation <valid-array.copy>`, a :ref:`value <syntax-val>` of :ref:`type <syntax-valtype>` :math:`\I32` is on the top of the stack.
+8. Pop the value :math:`s` from the stack.
```
rossberg, line 1137:
```suggestion
8. Pop the value :math:`\I32.\CONST~s` from the stack.
```

document/core/exec/instructions.rst
```diff
@@ -1120,75 +1120,67 @@ Reference Instructions
...
+9. Assert: due to :ref:`validation <valid-array.copy>`, a :ref:`value <syntax-val>` of :ref:`type <syntax-valtype>` :math:`(\REF~\NULL~y)` is on the top of the stack.
 
-11. Assert: due to :ref:`validation <valid-array.copy>`, a :ref:`value <syntax-val>` of :ref:`type <syntax-valtype>` :math:`(\REF~\NULL~y)` is on the top of the stack.
+10. Pop the value :math:`\reff_2` from the stack.
 
-12. Pop the value :math:`\reff_2` from the stack.
+11. Assert: due to :ref:`validation <valid-array.copy>`, a :ref:`value <syntax-val>` of :ref:`type <syntax-valtype>` :math:`\I32` is on the top of the stack.
 
-13. Assert: due to :ref:`validation <valid-array.copy>`, a :ref:`value <syntax-val>` of :ref:`type <syntax-valtype>` :math:`\I32` is on the top of the stack.
+12. Pop the value :math:`d` from the stack.
```
rossberg, line 1145:
```suggestion
12. Pop the value :math:`(\I32.\CONST~d)` from the stack.
```

document/core/syntax/instructions.rst
```diff
@@ -541,6 +541,15 @@ The instructions |I31NEW| and :math:`\I31GET\K{\_}\sx` convert between type |I31
...
 
+The following auxiliary function produces an instruction that accesses an array slot for a given :ref:`storage type <syntax-storagetype>`.
+
+.. _aux-getfield:
+
+.. math::
+   \begin{array}{lll}
+   \getfield(\valtype) &=& \ARRAYGET \\
+   \getfield(\packedtype) &=& \ARRAYGETU \\
+   \end{array}
```
rossberg, line 552: I'd move this next to the rules (as a "where:" sort of thing), since it is local helper for that specific rule only.

tlively:
Comments addressed. I also updated the formatting of the side conditions now that they are shorter.

rossberg (approved):
Thanks!

PR #420, opened 2023-09-09
Spec sequential visibility for globals

rossberg:
Also, add a couple of respective test cases.

### DIFF
--- a/document/core/valid/modules.rst
+++ b/document/core/valid/modules.rst
@@ -221,11 +221,14 @@ Memories :math:`\mem` are classified by :ref:`memory types <syntax-memtype>`.
    pair: validation; global
    single: abstract syntax; global
 .. _valid-global:
+.. _valid-globalseq:
 
 Globals
 ~~~~~~~
 
 Globals :math:`\global` are classified by :ref:`global types <syntax-globaltype>` of the form :math:`\mut~t`.
+
+Sequences of globals are handled incrementally, such that each definition has access to previous definitions.
 
 
 :math:`\{ \GTYPE~\mut~t, \GINIT~\expr \}`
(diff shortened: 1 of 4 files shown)


document/core/valid/modules.rst
```diff
@@ -251,6 +254,38 @@ Globals :math:`\global` are classified by :ref:`global types <syntax-globaltype>
...
+   ~\\
+   \frac{
+   }{
+     C \vdashglobalseq \epsilon : \epsilon
+   }
+   \qquad
+   \frac{
+     C \vdashglobal \global_1 : \X{gt}_1
+     \qquad
+     C, \CGLOBALS~\X{gt}_1 \vdashglobalseq \global^\ast : \X{gt}^\ast
```
tlively, line 282: Since context extension was originally defined with labels in mind, it prepends the new items to the relevant index space in the context, but here we want to append the new global to the end of the index space. It seems we need to define separate notations for prepending and appending extension, or else define a convention that extension prepends for labels and appends for everything else.

document/core/valid/modules.rst
```diff
@@ -802,8 +833,7 @@ The :ref:`external types <syntax-externtype>` classifying a module may contain f
    However, this recursion is just a specification device.
    All types needed to construct :math:`C` can easily be determined from a simple pre-pass over the module that does not perform any actual validation.
 
-   Globals, however, are not recursive and not accessible within :ref:`constant expressions <valid-const>` when they are defined locally.
-   The effect of defining the limited context :math:`C'` for validating certain definitions is that they can only access functions and imported globals and nothing else.
+   Globals, however, are not recursive but eveluated sequentially, such that each :ref:`constant expressions <valid-const>` only has access to imported or priviously defined globals.
```
tlively, line 836:
```suggestion
   Globals, however, are not recursive but evaluated sequentially, such that each :ref:`constant expressions <valid-const>` only has access to imported or previously defined globals.
```

tlively (commented):
LGTM beyond those comments.

document/core/valid/modules.rst
rossberg: Ah, yes, good catch. I'll probably redesign that at some point, but for now I'm using `\compose` instead here.

rossberg:
@tlively, PTAL.

tlively (approved):
Should we briefly describe what \compose means in the conventions section? Other than that, LGTM.

rossberg:
It's already described there. ;)

PR #421, opened 2023-09-09
[test] Add tests for packed struct fields

takikawa:
I noticed there weren't many tests for packed struct operations, and none for `struct.get_u`. This patch adds some more tests in `struct.wast` for those cases.


rossberg (approved):
Thanks!

PR #422, opened 2023-09-10
Rename i31.new mnemonic to ref.i31

rossberg:
This renames the `i31.new` mnemonic to `ref.i31`.

Rationale: We should reserve `new` for instructions that allocate a (fresh) object on the heap. The whole point of i31ref is that it does _not_ allocate anything, nor create a distinguished new reference. The instruction hence rather is in the same category as `ref.null` and `ref.func`.

Only affects the text format.


tlively:
Seems fine to me. @jakobkummerow, @eqrion, or anyone else, any objections?

jakobkummerow:
No objections, it's just a name.

tlively (approved):

PR #427, opened 2023-09-12
Rename text "deftype" to "typedef"

tlively:
Since a deftype is already something else outside the text format. Also fix an
abbreviation to correctly use "typedef" instead of "subtype". Adapted from a fix
by @bvisness in #413.


rossberg (approved):
Thanks for that!

PR #429, opened 2023-09-13
[spec] Add missing side conditions

rossberg:
Add side conditions on ref.test/ref.cast rules that check that the RHS type is well-formed before invoking subtyping.

@conrad-watt, PTAL

### DIFF
--- a/document/core/exec/instructions.rst
+++ b/document/core/exec/instructions.rst
@@ -294,9 +294,11 @@ Reference Instructions
 .. math::
    \begin{array}{lcl@{\qquad}l}
    \reff_1~\reff_2~\REFEQ &\stepto& (\I32.\CONST~1)
+     & (\iff \reff_1 = (\REFNULL~\X{ht}_1) \land \reff_2 = (\REFNULL~\X{ht}_2)) \\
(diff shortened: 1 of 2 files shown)


document/core/exec/instructions.rst
```diff
@@ -293,10 +293,12 @@ Reference Instructions
 
 .. math::
    \begin{array}{lcl@{\qquad}l}
+   \reff_1~\reff_2~\REFEQ &\stepto& (\I32.\CONST~1)
+     & (\iff \reff_1 = (\REFNULL~\X{ht}_1) \land \reff_2 = (\REFNULL~\X{ht}_2)) \\
```
conrad-watt, line 297: Just to check, this is definitely the semantics we want even though we have "typed" nulls?

rossberg: Yes, otherwise implementations would need to carry around type information with their null pointers.

"Told you so" — It's a wart that I already pointed out back when we made the dubious decision to dump subtyping for type-indexed nulls. ;)

conrad-watt (approved):
lgtm

PR #434, opened 2023-09-19
Fix context for table initialisers

### LINKED ISSUE #433: Section ordering issue in global.wast test
See https://github.com/WebAssembly/gc/blob/main/test/core/global.wast#L622.

```
(module
  ...
  (global $gn funcref (ref.null func))
  ...
  (table $t 10 funcref (global.get $gn))
)
```

This module is supposed to be valid, and has a table with a default value expression referencing a global. However the global section comes after the table section, so when validating the global.get in the table, `$gn` has not been decoded yet.

Also, it looks like the text format for table default value initializers is not written yet.

rossberg:
Fix #433.

### DIFF
--- a/document/core/valid/modules.rst
+++ b/document/core/valid/modules.rst
@@ -719,16 +719,17 @@ The :ref:`external types <syntax-externtype>` classifying a module may contain f
 
   * all other fields are empty.
 
-* Under the context :math:`C'`,
-  the sequence :math:`\module.\MGLOBALS` of :ref:`globals <syntax-global>` must be :ref:`valid <valid-globalseq>` with a sequence :math:`\X{gt}^\ast` of :ref:`global types <syntax-globaltype>`.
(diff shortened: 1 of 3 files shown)


titzer (approved):
lgtm.

I guess the longer term solution should be to relax section order to allow some globals to be declared before tables?

rossberg:
Yes, absolutely. And for that to work cleanly in the future, we must not mess up the dependency order now.

I'd love to do the relaxation of section order and repetition sooner rather than later, but despite many weak use cases, we are still missing a strong one, it seems.

titzer:
Also, should it be legal to use an imported global? Imports occur before tables and thus add to the global index space, so it seems to me like this should be legal.

rossberg:
@titzer, that is legal, but it seems we were missing a test for that. PTAL at WebAssembly/function-references#106

PR #438, opened 2023-09-20
List possible implementation limits

rossberg:
Noticed that we still needed to extend those bullet lists in the appendix.

### DIFF
--- a/document/core/appendix/implementation.rst
+++ b/document/core/appendix/implementation.rst
@@ -81,6 +81,7 @@ For a module given in :ref:`binary format <binary>`, additional limitations may 
 
 * the size of a :ref:`module <binary-module>`
 * the size of any :ref:`section <binary-section>`
+* the size of an individual :ref:`constant <valid-constant>` :ref:`expression <syntax-expr>`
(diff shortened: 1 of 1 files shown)


document/core/appendix/implementation.rst
```diff
@@ -50,7 +53,8 @@ An implementation may impose restrictions on the following dimensions of a modul
 * the size of a :ref:`structured control instruction <syntax-instr-control>`
 * the number of :ref:`structured control instructions <syntax-instr-control>` in a :ref:`function <syntax-func>`
 * the nesting depth of :ref:`structured control instructions <syntax-instr-control>`
-* the number of :ref:`label indices <syntax-labelidx>` in a |brtable| instruction
+* the number of :ref:`label indices <syntax-labelidx>` in a |BRTABLE| instruction
+* the length of the array in a |ARRAYNEWFIXED| instruction
```
titzer, line 57: AFAICT this is also limited by the bytecode length, but I'm fine specifying it separately this way.

Do we have a limitation on the size of an initializer expression?

rossberg: Good point, added.

PR #439, opened 2023-09-20
[test] Add boundary cast tests for ToWebAssemblyValue

takikawa:
This PR adds JS API tests for exercising `ToWebAssemblyValue` with all the categories of GC types. It adds on top of the previous two JS API tests PRs, so it should be merged after them (just the last commit is new).

With these three tests, I think we will have good test coverage of the JS API, but if anyone sees any missing cases that should be covered that would be helpful to know.

The tests pass on JSC (w/ some unlanded patches), V8, and SpiderMonkey (w/ final opcodes) with one exception that I'll comment below.

### DIFF
--- a/test/js-api/gc/casts.tentative.any.js
+++ b/test/js-api/gc/casts.tentative.any.js
@@ -1,0 +1,332 @@
+// META: global=window,dedicatedworker,jsshell
+// META: script=/wasm/jsapi/wasm-module-builder.js
+
+let exports = {};
+setup(() => {
+  const builder = new WasmModuleBuilder();
+  const structIndex = builder.addStruct([makeField(kWasmI32, true)]);
+  const arrayIndex = builder.addArray(kWasmI32, true);
+  const structIndex2 = builder.addStruct([makeField(kWasmF32, true)]);
+  const arrayIndex2 = builder.addArray(kWasmF32, true);
+  const funcIndex = builder.addType({ params: [], results: [] });
+  const funcIndex2 = builder.addType({ params: [], results: [kWasmI32] });
+
+  const argFunctions = [
+    { name: "any", code: kWasmAnyRef },
+    { name: "eq", code: kWasmEqRef },
+    { name: "struct", code: kWasmStructRef },
+    { name: "array", code: kWasmArrayRef },
+    { name: "i31", code: kWasmI31Ref },
+    { name: "func", code: kWasmFuncRef },
+    { name: "extern", code: kWasmExternRef },
+    { name: "none", code: kWasmNullRef },
+    { name: "nofunc", code: kWasmNullFuncRef },
+    { name: "noextern", code: kWasmNullExternRef },
+    { name: "concreteStruct", code: structIndex },
+    { name: "concreteArray", code: arrayIndex },
+    { name: "concreteFunc", code: funcIndex },
+  ];
+
+  for (const desc of argFunctions) {
+    builder
+      .addFunction(desc.name + "Arg", makeSig_v_x(wasmRefType(desc.code)))
+      .addBody([])
+      .exportFunc();
+
+    builder
+      .addFunction(desc.name + "NullableArg", makeSig_v_x(wasmRefNullType(desc.code)))
+      .addBody([])
+      .exportFunc();
+  }
+
+  builder
+    .addFunction("makeStruct", makeSig_r_v(wasmRefType(structIndex)))
+    .addBody([...wasmI32Const(42),
+              ...GCInstr(kExprStructNew), structIndex])
+    .exportFunc();
+
+  builder
+    .addFunction("makeArray", makeSig_r_v(wasmRefType(arrayIndex)))
+    .addBody([...wasmI32Const(5), ...wasmI32Const(42),
+              ...GCInstr(kExprArrayNew), arrayIndex])
+    .exportFunc();
+
+  builder
+    .addFunction("makeStruct2", makeSig_r_v(wasmRefType(structIndex2)))
+    .addBody([...wasmF32Const(42),
+              ...GCInstr(kExprStructNew), structIndex2])
+    .exportFunc();
+
+  builder
+    .addFunction("makeArray2", makeSig_r_v(wasmRefType(arrayIndex2)))
+    .addBody([...wasmF32Const(42), ...wasmI32Const(5),
+              ...GCInstr(kExprArrayNew), arrayIndex2])
+    .exportFunc();
+
+  builder
+    .addFunction("testFunc", funcIndex)
+    .addBody([])
+    .exportFunc();
+
+  builder
+    .addFunction("testFunc2", funcIndex2)
+    .addBody([...wasmI32Const(42)])
+    .exportFunc();
+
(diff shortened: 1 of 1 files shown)


test/js-api/gc/casts.tentative.any.js
```diff
@@ -0,0 +1,200 @@
...
+
+test(() => {
+  exports.anyArg(exports.makeStruct());
+  exports.anyArg(exports.makeArray());
+  exports.anyArg(42);
+  exports.anyArg(42n);
+  exports.anyArg("foo");
+  exports.anyArg({});
+  exports.anyArg(() => {});
+  assert_throws_js(TypeError, () => exports.anyArg(exports.testFunc));
```
takikawa, line 108: This line of the anyref test fails on V8 and SpiderMonkey because they do not throw. If I'm reading the JS API spec correctly though, `ToWebAssemblyValue` should throw here.
 
The relevant steps of the algorithm are from step 6:

> 6. If type is of the form [ref](https://webassembly.github.io/spec/core/syntax/types.html#syntax-reftype) null heaptype,
> [...]
>    4. Else if v is an [Exported Function](https://webassembly.github.io/gc/js-api/index.html#exported-function),
>      1. Let funcaddr be the value of v’s [[FunctionAddress]] internal slot.
>      2. Let r be [ref.func](https://webassembly.github.io/spec/core/exec/runtime.html#values) funcaddr.

In this case, `exports.testFunc` should be an exported function and this step will match. Thus `r` will be a `ref.func <addr>` and its type will be a typed funcref.

> 8. Let actualtype be [ref_type](https://webassembly.github.io/spec/core/appendix/embedding.html#embed-ref-type)(store, r).
> 9. If [match_valtype](https://webassembly.github.io/spec/core/appendix/embedding.html#embed-match-valtype)(actualtype, type) is false,
>    1. Throw a [TypeError](https://tc39.github.io/ecma262/#sec-native-error-types-used-in-this-standard-typeerror).

Here `actualtype` should be some function type (is it specified that `ref_type` must return the most specific type?). `match_type(<functype>, (ref any))` will return false because functions cannot be a subtype of any. Thus this should throw.

Does this seem like the correct reading of the spec?

rossberg: Yes, that's the correct reading. Whether it's actually doing what's intended is a different question. :)  Given that any JS value is in externref, and every externref can be converted to anyref, we may want to allow all JS values for any. For that, we'd need to add a special case under step 9, that if heaptype is any, instead of throwing, construct a ref.host (as in step 6).

(Yes, ref_type returns the most specific type.)

takikawa: Ok, that makes sense to me given that wasm's type system will keep exported functions as hostrefs and funcrefs separate. I can change this test to not throw, in which case the browsers will all agree (for JSC the patch for this behavior hasn't landed, so I can just change it).

For the special case, it seems like it would be equivalent to say in step 4:

> Else if v is an [Exported Function](https://webassembly.github.io/gc/js-api/index.html#exported-function) and heaptype != any

in which case it falls through to the hostref case (so that it's not necessary to repeat the ref.host construction prose).

rossberg: Sounds good, please see #440.

takikawa: I just updated this test case to match the new spec PR.

test/js-api/gc/casts.tentative.any.js
```diff
@@ -0,0 +1,303 @@
...
+  assert_throws_js(TypeError, () => exports.funcNullableArg(42));
+  assert_throws_js(TypeError, () => exports.funcNullableArg(42n));
+  assert_throws_js(TypeError, () => exports.funcNullableArg("foo"));
+  assert_throws_js(TypeError, () => exports.funcNullableArg({}));
+  assert_throws_js(TypeError, () => exports.funcNullableArg(() => {}));
+}, "funcref casts");
+
+test(() => {
+  for (const nullfunc of [exports.noneArg, exports.nofuncArg, exports.noexternArg]) {
+    assert_throws_js(TypeError, () => nullfunc(exports.makeStruct()));
```
takikawa, line 209: I added a few more test cases as I realized tests for bottom types were missing.

Here, V8 (commit 014e0bb5ac4691c85b1da2dac579af77d02d17b7) does not throw while Spidermonkey and JSC do. This test is checking all of the bottom types, and in V8 the failure is just for `noextern`.

I believe casting to `noextern` should fail for anything but null (if it's a nullable ref).

takikawa: @jakobkummerow could the above be a V8 bug?

rossberg: > I believe casting to noextern should fail for any of the arguments but null (if it's a nullable ref, as in the loop below this one).

I think you're right, but the spec is wrong about that: on the extern ref path, it skips the subtype check, which is wrong. Please see #442 for a fix.

takikawa:
I'm going to be out of office for two weeks but if anyone has feedback about this or any of the other JS API PRs I can take a look and address them once I'm back.

test/js-api/gc/casts.tentative.any.js
```diff
@@ -0,0 +1,303 @@
...
+  assert_throws_js(TypeError, () => exports.concreteArrayNullableArg(42));
+  assert_throws_js(TypeError, () => exports.concreteArrayNullableArg(42n));
+  assert_throws_js(TypeError, () => exports.concreteArrayNullableArg("foo"));
+  assert_throws_js(TypeError, () => exports.concreteArrayNullableArg({}));
+  assert_throws_js(TypeError, () => exports.concreteArrayNullableArg(exports.testFunc));
+  assert_throws_js(TypeError, () => exports.concreteArrayNullableArg(() => {}));
+}, "concrete array casts");
+
+test(() => {
+  exports.concreteFuncArg(exports.testFunc);
```
rossberg, line 282: Maybe have another one with a testFunc2 here and below, analogous to structs and arrays.

test/js-api/gc/casts.tentative.any.js
```diff
@@ -0,0 +1,303 @@
...
+    { name: "struct", code: kWasmStructRef },
+    { name: "array", code: kWasmArrayRef },
+    { name: "i31", code: kWasmI31Ref },
+    { name: "func", code: kWasmFuncRef },
+    { name: "none", code: kWasmNullRef },
+    { name: "nofunc", code: kWasmNullFuncRef },
+    { name: "noextern", code: kWasmNullExternRef },
+    { name: "concreteStruct", code: structIndex },
+    { name: "concreteArray", code: arrayIndex },
+    { name: "concreteFunc", code: funcIndex },
```
rossberg, line 25: The one type I'm missing in this list is `extern`.

takikawa:
@rossberg I think the latest commit should address your comments.

There are merge conflicts but I can rebase to fix once the PR looks ok.

rossberg (approved):

takikawa:
This is now rebased and should merge cleanly.

PR #440, opened 2023-09-21
[js-api] Allow exported funcs as any

rossberg:
As per [discussion](https://github.com/WebAssembly/gc/pull/439#discussion_r1332282446).

@takikawa, PTAL. Reordered slightly to make it clearer.


takikawa (approved):
LGTM

PR #442, opened 2023-09-21
[js-api] Fix type check for externrefs

rossberg:
See [comment](https://github.com/WebAssembly/gc/pull/439/files#r1333519677). @takikawa, PTAL.

### DIFF
--- a/document/js-api/index.bs
+++ b/document/js-api/index.bs
@@ -1147,11 +1147,14 @@ The algorithm <dfn>ToWebAssemblyValue</dfn>(|v|, |type|) coerces a JavaScript va
         1. Let |f64| be |number|.
     1. Return [=f64.const=] |f64|.
 1. If |type| is of the form [=ref=] |null| |heaptype|,
-    1. If [=match_valtype=](|type|, [=ref=] |null| [=heap-type/extern=]),
-        1. Let |rr| be [=ToWebAssemblyValue=](|v|, [=ref=] |null| [=heap-type/any=]).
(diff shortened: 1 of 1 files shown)


document/js-api/index.bs
```diff
@@ -1148,9 +1148,9 @@ The algorithm <dfn>ToWebAssemblyValue</dfn>(|v|, |type|) coerces a JavaScript va
     1. Return [=f64.const=] |f64|.
 1. If |type| is of the form [=ref=] |null| |heaptype|,
     1. If [=match_valtype=](|type|, [=ref=] |null| [=heap-type/extern=]),
-        1. Let |r| be [=ToWebAssemblyValue=](|v|, [=ref=] |null| [=heap-type/any=]).
-        1. Return [=ref.extern=] |r|.
-    1. If |v| is null,
+        1. Let |rr| be [=ToWebAssemblyValue=](|v|, [=ref=] |null| [=heap-type/any=]).
+        1. Let |r| be [=ref.extern=] |rr|.
+    1. Else if |v| is null,
```
takikawa, line 1153: I think the null case and extern case may need to be swapped. If `v` is null and `type` is `(ref null noextern)` then this current conditional will pick the extern branch and set `r = ref.extern (ref.null noextern)` which has type `extern`. Then the subtype check fails.

rossberg: Yes, you're right. I also moved the func case back to the beginning, since it is more similar to extern now.

takikawa (approved):
LGTM

PR #447, opened 2023-10-05
Rename extern.in/externalize

### LINKED ISSUE #432: Rename extern.internalize/extern.externalize?
In discussions around adding an 'exnref' type (https://github.com/WebAssembly/exception-handling/issues/280), it seemed like there was desire for it to be a new type hierarchy, like any and extern. If this is the case, it could be useful to have instructions to convert from 'exnref' to 'externref' or 'anyref'. This has sort of come up before (https://github.com/WebAssembly/exception-handling/issues/202).

I can't think of a good name for these though, as 'extern.internalize/externalize' seem to have a builtin assumption that you just want to convert between those two hierarchies.

What would folks think about renaming to: `any.from_extern` and `extern.from_any`? This would allow a natural `any.from_exn` or `extern.from_exn`?

rossberg:
Fix #432.


tlively (approved):
Thanks!

PR #451, opened 2023-10-16
[spec] Fix the instruction types of `array.new_default` and `array.len`

q82419:
According to the [validation of `array.new_default`](https://webassembly.github.io/gc/core/valid/instructions.html#xref-syntax-instructions-syntax-instr-array-mathsf-array-new-default-x), the type is `[i32] -> [(ref x)]`, but in the appendix page is `[] -> [(ref x)]`.

According to the [validation of `array.len`](https://webassembly.github.io/gc/core/valid/instructions.html#xref-syntax-instructions-syntax-instr-array-mathsf-array-len), the type is `[arrayref] -> [i32]`, but in the appendix page is `[arrayref] -> []`.


rossberg (approved):
Thanks!

PR #455, opened 2023-10-20
[test] Prevent from module name conflict when registering in wast.

q82419:
Such as the `register` commands in [linking.wast](https://github.com/WebAssembly/gc/blob/main/test/core/linking.wast), I think it's better to prevent from the module name conflict for each `register` commands in a `wast` file.


rossberg (approved):
Makes sense, thanks.

PR #458, opened 2023-10-29
[spec] Fix typing of functions

rossberg:
Adjust various places that were still using a functype or a typeidx instead of a deftype to type functions. Also, add missing subsumtion for external values and reference values.

@q82419, please take a look.


q82419:
Looks excellent, thanks for help!

PR #462, opened 2023-10-30
[spec] fix immediate of `array.set` instruction.

q82419:
The `array.set` has no `i` immediate.

### DIFF
--- a/document/core/valid/instructions.rst
+++ b/document/core/valid/instructions.rst
@@ -566,8 +566,8 @@ Aggregate Reference Instructions
 
 .. _valid-array.set:
 
-:math:`\ARRAYSET~x~i`
-.....................
+:math:`\ARRAYSET~x`
+...................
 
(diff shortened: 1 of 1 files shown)


document/core/valid/instructions.rst
```diff
@@ -566,7 +566,7 @@ Aggregate Reference Instructions
 
 .. _valid-array.set:
 
-:math:`\ARRAYSET~x~i`
+:math:`\ARRAYSET~x`
 .....................
```
rossberg, line 570: Can you also adjust the underline? Otherwise Sphinx will complain.

rossberg (commented):
Thanks!

document/core/valid/instructions.rst
q82419: Oops, forgot it. Thanks.

rossberg (approved):

PR #463, opened 2023-11-03
Pass index argument to `getfield` function

ShinWonho:
As far as I know, `array.get` instruction requires a type index, so type index should be passed to `getfield` function.

Screenshot 2023-11-03 at 1 57 00 PM

* Current specification

Screenshot 2023-11-03 at 2 02 53 PM

* Suggested change

### DIFF
--- a/document/core/exec/instructions.rst
+++ b/document/core/exec/instructions.rst
@@ -1153,7 +1153,7 @@ Reference Instructions
 
     d. Push the value :math:`\I32.\CONST~s` to the stack.
 
-    e. Execute :math:`\getfield(\X{st})`.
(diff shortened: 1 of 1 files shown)


rossberg (commented):
Ah, thanks!

But note that this wants the source type, so it ought to be `y`. (There was a similar bug in spectec wasm3, which I just fixed.)

Also, getfield is sort of a local definition in a "where" clause, so we can assume that it can use `y` directly, it doesn't need to be passed as a parameter.

ShinWonho:
Screenshot 2023-11-03 at 5 36 10 PM

Thanks for the review!
Now, I fixed it.

rossberg (approved):
Thanks!

PR #464, opened 2023-11-03
Fix `array.get/set` reduction rule

ShinWonho:
Trap case of `array.get` and `array.set` is omitted or wrongly described.

Screenshot 2023-11-03 at 2 29 04 PM
Screenshot 2023-11-03 at 2 29 29 PM

* Current specification

Screenshot 2023-11-03 at 2 28 46 PM
Screenshot 2023-11-03 at 2 30 09 PM

* Suggested change


rossberg (approved):
Indeed, thanks!

rossberg (approved):
Indeed, thanks!

PR #466, opened 2023-11-03
[test] Add `assert_return`

f52985:
This PR adds `assert_return` around the invocations of two test functions at the end of `gc/ref_test.wast`

### DIFF
--- a/test/core/gc/ref_test.wast
+++ b/test/core/gc/ref_test.wast
@@ -326,5 +326,5 @@
   )
 )
 
-(invoke "test-sub")
-(invoke "test-canon")
+(assert_return (invoke "test-sub"))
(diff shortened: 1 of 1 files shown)


rossberg (approved):
I'm okay either way, but the change doesn't add anything relevant to the test. The test verifies the absence of a trap, and a trap will be recognised as a test failure regardless of the assertion.

f52985:
Oh, I see. I thought the result of `invoke` only becomes relevant when coupled with assertions, and an `invoke` without an assertion as a side-effect generator.

Still, I think it would be slightly better to clearly indicate that the purpose of these invocations are actually "asserting" if the program showed the intended behavior.

rossberg:
I'm fine with adding it.

Issue #461, opened 2023-10-30
JS API: importing globals


vouillon:
I'm wondering whether the ['read the imports' process](https://webassembly.github.io/gc/js-api/#read-the-imports) should be updated. As far as I can see, when importing a global (step 3.5), the imported value should be either a number or should implement `Global`. So, it currently does not allow to directly import a JavaScript string, for instance.

It seems that, when the global has a reference type, V8 allows to import arbitrary values as long as the type matches. Firefox is currently pickier but allows to import arbitrary values when the global has type `externref`.

rossberg:
Good point, I agree it needs some updating to account for reference types. And corresponding tests.

@takikawa, can you look into this?

rossberg:
Closed via #467.

PR #467, opened 2023-11-04
[js-api] Update the 'read the imports' process

vouillon:
Handle the case where the global has a reference type. See #461.


rossberg (approved):
Thanks, LGTM.

PR #473, opened 2023-11-07
fix rule for array.set

zapashcanon:
Hi,

In the `array.set` rule, there is `if expand(F.module.types[x]) = struct ft^n` but it should be `if expand(F.module.types[x]) = array ft`.


rossberg (approved):
Indeed, thanks!

PR #474, opened 2023-11-08
remove duplicated token definition in parser

zapashcanon:
It is already defined two lines above. This was introduced in #422.

`ocamlyacc` doesn't produce an error in these cases, but it probably should. I opened [an issue upstream](https://github.com/ocaml/ocaml/issues/12723) .

I tried on the Owi interpreter where I'm using Menhir and it is producing an error when this is the case: `Error: the token REF_I31 has multiple definitions.`.

It would be nice to switch to Menhir at some point (which is the standard in OCaml nowadays) but I guess introducing one more dependency is not acceptable ?


rossberg (approved):

zapashcanon:
Thanks for merging @rossberg. What are your thoughts about switching to Menhir ? As stated in the issue linked above, `ocamlyacc` is almost considered as deprecated. I'm willing to send a PR if we can accept Menhir as a dependency.

rossberg:
I'd be fine with it.

PR #478, opened 2023-11-10
Fix ref.null semantics

ShinWonho:
### Problem
Screenshot 2023-11-10 at 5 47 05 PM

This is the semantics of `ref.null ht` in the current specification.
But I think simply pushing `ref.null ht` is problematic.

### Example

Consider the following example, which is simplified version of "test-sub" in ref_test.wast.

```
(module
  (type $t0 (sub (struct)))

  (func (export "test-sub")
    ;; must hold
    (ref.test (ref null $t0) (ref.null $t0))
    (drop)))

(assert_return (invoke "test-sub"))
```

Screenshot 2023-11-10 at 6 08 14 PM

According to the semantics of `ref.test`, we first need to get the reference type of `ref.null $t0`.

Screenshot 2023-11-10 at 6 10 50 PM

However, the typeindex `$t0` is not ok under empty context.

### Suggested change

In the reference interpreter, type index of `ref.null $t0` is substituted before pushed.

```
let rec step (c : config) : config =

      ....

      | RefNull t, vs' ->
        Ref (NullRef (subst_heap_type (subst_of c.frame.inst) t)) :: vs', []
```

Therefore, I fixed the semantics according to the reference interpreter.

Screenshot 2023-11-10 at 5 46 24 PM

### DIFF
--- a/document/core/exec/instructions.rst
+++ b/document/core/exec/instructions.rst
@@ -192,25 +192,21 @@ Reference Instructions
 
 .. _exec-ref.null:
 
-:math:`\REFNULL~\X{ht}`
+:math:`\REFNULL~x`
 .......................
 
 1. Let :math:`F` be the :ref:`current <exec-notation-textual>` :ref:`frame <syntax-frame>`.
 
-2. If :math:`\X{ht}` is :math:`\typeidx`, then:
+2. Assert: due to :ref:`validation <valid-ref.null>`, the :ref:`defined type <syntax-deftype>` :math:`F.\AMODULE.\MITYPES[x]` exists.
 
-  a. Let :math:`\X{ht}_1` be the :ref:`heap type <syntax-heaptype>` :math:`\insttype_{F.\AMODULE}(\X{ht})`.
+3. Let :math:`\deftype` be the :ref:`defined type <syntax-deftype>` :math:`F.\AMODULE.\MITYPES[x]`.
 
-  b. Push the value :math:`\REFNULL~\X{ht}_1` to the stack.
-
-3. Else:
-
-  a. Push the value :math:`\REFNULL~\X{ht}` to the stack.
+4. Push the value :math:`\REFNULL~\deftype` to the stack.
 
 .. math::
    \begin{array}{lcl@{\qquad}l}
-   F; (\REFNULL~x) &\stepto& F; (\REFNULL~\X{ht})
-     & (\iff \X{ht} = \insttype_{F.\AMODULE}(\X{x})) \\
+   F; (\REFNULL~x) &\stepto& F; (\REFNULL~\deftype)
+     & (\iff \deftype = F.\AMODULE.\MITYPES[x]) \\
    \end{array}
 
 
--- a/document/core/exec/instructions.rst
+++ b/document/core/exec/instructions.rst
@@ -210,9 +210,8 @@ Reference Instructions
    \end{array}
 
 .. note::
-   This is the reduction rule for |REFNULL| |TYPEIDX|.
-   No formal reduction rule is required for |REFNULL| |ABSHEAPTYPE|,
-   since the instruction is already a :ref:`value <syntax-val>`.
+   No formal reduction rule is required for the case |REFNULL| |ABSHEAPTYPE|,
+   since the instruction form is already a :ref:`value <syntax-val>`.
 
 
 .. _exec-ref.func:
--- a/document/core/exec/instructions.rst
+++ b/document/core/exec/instructions.rst
@@ -195,10 +195,23 @@ Reference Instructions
 :math:`\REFNULL~\X{ht}`
 .......................
 
-1. Push the value :math:`\REFNULL~\X{ht}` to the stack.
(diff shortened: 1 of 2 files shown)


rossberg (commented):
Ah, good catch! This is rather annoying, if only ref.null didn't have this dreaded type annotation...

Unfortunately, I think the fix isn't quite as easy as that. Because (ref.null ht) is defined to be a value, and values must never reduce, otherwise the Progress statement is compromised.

So we need to syntactically distinguish reduced null refs from non-reduced ones. We could do it based on the shape of the ht, which is probably fine for now (you'll need to add a side condition to the grammar of reference values), but is a bit hacky and not very robust.

Long-term, I think the cleaner fix is to make the type annotation optional. Then the reduction rule would remove it. And the typing rule for valid-ref would only allow unannotated ref.null and assign it type (ref null bot).

But changing the syntax of reference values like that requires more adjustments throughout the spec, and even the JS API spec. The interpreter should be adjusted as well then. That's quite a bit of work, so perhaps for later.

zapashcanon:
> Because (ref.null ht) is defined to be a value, and values must never reduce, otherwise the Progress statement is compromised.

Wouldn't a simple fix to have `null.new $t` be an instruction and `ref.null $t` a value ?

rossberg:
See my reply on the other issue. Also, ref.null is already standardised as part of Wasm 2.0 (as is ref.func, btw), so we cannot change it anymore.

ShinWonho:
Screenshot 2023-11-13 at 5 33 26 PM

Screenshot 2023-11-13 at 5 33 56 PM

I changed the syntax of `ref.null` value so that `ref.null` with type index cannot be a value.
Is this right fix?

document/core/exec/runtime.rst
```diff
@@ -47,7 +47,7 @@ Any of the aformentioned references can furthermore be wrapped up as an *externa
    \production{vector} & \vecc &::=&
      \V128.\CONST~\i128 \\
    \production{reference} & \reff &::=&
-     \REFNULL~t \\&&|&
+     \REFNULL~t & (\iff t \neq \typeidx) \\&&|&
```
rossberg, line 50: This is correct in spirit, but technically, since meta variables are existentially quantified, a negation like this doesn't do the right thing – you can always take a different type index to make it true. I think you need to say some like "t = absheaptype ∨ t = deftype". Or perhaps spell it out as `\REFNULL~\absheaptype~|~\REFNULL~\deftype`.

ShinWonho: Thank you for the comment! I fixed this.

document/core/exec/instructions.rst
```diff
@@ -192,13 +192,22 @@ Reference Instructions
...
 
-:math:`\REFNULL~\X{ht}`
+:math:`\REFNULL~x`
 .......................
 
-1. Push the value :math:`\REFNULL~\X{ht}` to the stack.
+1. Let :math:`F` be the :ref:`current <exec-notation-textual>` :ref:`frame <syntax-frame>`.
 
-.. note::
-   No formal reduction rule is required for this instruction, since the |REFNULL| instruction is already a :ref:`value <syntax-val>`.
```
rossberg, line 201: I think a variation of this note still applies, since we only reduce _some_ ref.null's.

document/core/exec/runtime.rst
```diff
@@ -47,7 +47,8 @@ Any of the aformentioned references can furthermore be wrapped up as an *externa
    \production{vector} & \vecc &::=&
      \V128.\CONST~\i128 \\
    \production{reference} & \reff &::=&
-     \REFNULL~t \\&&|&
+     \REFNULL~\absheaptype \\&&|&
+     \REFNULL~\deftype \\&&|&
```
rossberg, line 51: In fact, I wouldn't mind:
```suggestion
     \REFNULL~(\absheaptype~|~\deftype) \\&&|&
```

document/core/exec/instructions.rst
ShinWonho: Screenshot 2023-11-14 at 8 13 01 PM
I added a note like this. If you think there are better words, feel free to change it.

document/core/exec/instructions.rst
```diff
@@ -192,13 +192,27 @@ Reference Instructions
...
+   \begin{array}{lcl@{\qquad}l}
+   F; (\REFNULL~x) &\stepto& F; (\REFNULL~\deftype)
+     & (\iff \deftype = F.\AMODULE.\MITYPES[x]) \\
+   \end{array}
 
 .. note::
-   No formal reduction rule is required for this instruction, since the |REFNULL| instruction is already a :ref:`value <syntax-val>`.
+   This is the reduction rule for |REFNULL| |TYPEIDX|.
+   No formal reduction rule is required for |REFNULL| |ABSHEAPTYPE|,
+   since the instruction is already a :ref:`value <syntax-val>`.
```
rossberg, line 215:
```suggestion
   No formal reduction rule is required for the case |REFNULL| |ABSHEAPTYPE|,
   since the instruction form is already a :ref:`value <syntax-val>`.
```

rossberg (approved):
Thanks!

PR #479, opened 2023-11-10
[spec] Tweak dimensionality in array.new_data

### LINKED ISSUE #476: Dimension mismatch in the premise of `array.new_data` reduction rule
image

I believe there is a dimension mismatch in the last premise of the above reduction rule.

The lhs, `(bytes_ft(c))^n`, is a 2d list of bytes, for 1d list of bytes, `bytes_ft(c)` is iterated by `^n`.
The rhs, `S.datas[F.module.dataaddrs[y]].data[s : n * |ft| / 8]`, on the other hand, is a 1d list of bytes.

It may be true that the lhs represents a 1d list of bytes, if we interpret `^n` as a concatenation of 1d lists.
But I do not think that it would be consistent with the other uses of `^n` throughout the entire document.

rossberg:
@jaehyun1ee, please take a look. Fix #476.


jaehyun1ee:
LGTM! Thank you for the edit :)

PR #480, opened 2023-11-10
Fix Overview.md inconsistencies

dylanahsmith:
A couple of the inconsistencies were due to changes in the dependent function-references proposal:
* `optref` seems to have been changed to `ref null` at some point
* the `let` instruction was removed in favour of validation (https://github.com/WebAssembly/function-references/pull/63)

The other couple of inconsistencies just seemed to accidentally referencing the wrong label in an example and description of an example.

### DIFF
--- a/proposals/gc/Overview.md
+++ b/proposals/gc/Overview.md
@@ -168,7 +168,7 @@ The above could map to
 )
 ```
 These functions `$f` and `$g` code introduces local with the `let` instruction (see the [typed function references proposal](https://github.com/WebAssembly/function-references/blob/master/proposals/function-references/Overview.md)) because the defined types cannot be null, such that locals of these types cannot be default-initialised.
-In the case of `$h` the local is declared as nullable, however, mapping to an optional reference.
+In the case of `$b` the local is declared as nullable, however, mapping to an optional reference.
 The respective access via `struct.get` may hence trap.
 
 
--- a/proposals/gc/Overview.md
+++ b/proposals/gc/Overview.md
@@ -153,7 +153,7 @@ The above could map to
 )
 
 (func $h
-  (local $b (optref $buf))
+  (local $b (ref null $buf))
   (local.set $b
     (struct.new $buf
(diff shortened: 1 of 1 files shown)


proposals/gc/Overview.md
```diff
@@ -408,7 +411,7 @@ For example:
 )
 ```
 All accesses are type-checked at validation time.
-The structure operand of `struct.get/set` may either be a `ref` or an `optref` for a structure type
+The structure operand of `struct.get/set` may either be a `ref` or an `ref null` for a structure type
```
rossberg, line 414:
```suggestion
The structure operand of `struct.get/set` may either be a `ref` or a `ref null` for a structure type
```

proposals/gc/Overview.md
```diff
@@ -440,7 +443,7 @@ Elements are accessed with generic load/store instructions that take a reference
 )
 ```
 The element type of every access is checked at validation time.
-The array operand of `array.get/set` may either be a `ref` or an `optref` for an array type
+The array operand of `array.get/set` may either be a `ref` or an `ref null` for an array type
```
rossberg, line 446:
```suggestion
The array operand of `array.get/set` may either be a `ref` or a `ref null` for an array type
```

proposals/gc/Overview.md
```diff
@@ -505,7 +508,7 @@ These notions are already introduced by [typed function references](https://gith
 
 Plain references cannot be null,
 avoiding any runtime overhead for null checks when accessing a struct or array.
-Nullable references are available as separate types called `optref`, as per the .
+Nullable references are available as separate types called `ref null`, as per the .
```
rossberg, line 511:
```suggestion
Nullable references are available as separate types called `ref null`, as per the [typed function references proposal](https://github.com/WebAssembly/function-references/blob/master/proposals/function-references/Overview.md).
```

rossberg (commented):
Thanks!

PR #483, opened 2023-11-10
Tweak note about global.get of locally-defined globals and constness

fitzgen:
Element and data segments are validated under the full `C` now, not the `C'` that only has imported globals, and therefore `global.get $local_global` is a valid constant expression for segments now.

@rossberg PTAL


rossberg:
Thanks, this PR was racing with #481, which I just merged, so I'm closing this.

fitzgen:
Ah, thanks, I should have seen that PR.

PR #484, opened 2023-11-13
Add current frame

ShinWonho:
Many prose related to struct and array omit "Let F be ..."
Not only that, `Array.copy` also omit F in the reduction rule.


rossberg (approved):
Thanks! Aside from array.copy, I assume this is just adding the first step everywhere and renumbers the others? It's a bit hard to see from the diff, because Github presents it so badly.

PR #504, opened 2023-12-26
[test] Unify the error message in `type-subtyping.wast`.

q82419:
`indirect call` -> `indirect call type mismatch`
`cast` -> `cast failure`
`incompatible import` -> `incompatible import type`


rossberg (approved):
I'm fine changing these, but wanted to point out that this and other message short forms were somewhat intentional, meant to make tests more robust with respect to fine-tuning of error messages in the interpreter.

PR #509, opened 2024-01-16
[spec] Fix prose for bounds check in array.init_data/elem

### LINKED ISSUE #508: [spec] array.init_elem: bounds check correct?
The bounds check in the spec for `array.init_elem` currently has the following text with "larger than or equal to". Should it be "larger than" instead?

image

If `d` (or `s`) and `n` are both zero, then this traps if the array is empty (or the segment is).

I don't think this is the intended behavior since `array.new_elem` checks "larger than" and the tests also don't trap in this case (e.g., [this test case](https://github.com/WebAssembly/gc/blob/main/test/core/gc/array_init_elem.wast#L107)).

rossberg:
Fix #508. Also tweak to be more symmetric to table/memory.init.


tlively (commented):
lgtm

PR #513, opened 2024-01-24
[spec] Add note and algorithmic rules about typing with top type

rossberg:
Address question in #511.


tlively:
cc @fitzgen, I believe this new note would have helped you as well.

conrad-watt (approved):
lgtm

fitzgen:
> cc @fitzgen, I believe this new note would have helped you as well.

Yes, it would have. Glad to see it added for future readers!

Issue #516, opened 2024-02-06
Does the combo of subtyping and unreachable code require actual type inference?


fitzgen:
Consider the following example. I've annotated it with the operand stack in between each opcode twice:

1. Once with the appendix's algorithm.
2. A second time with `wasm-smith`'s "actual" stack.

The difference is that the appendix algorithm truncates the stack after unreachable code, letting subsequent pops materialize their expected types, while `wasm-smith` maintains the "actual" stack going into the unreachable code.

AFAICT, the formal typing rules don't actually mandate the appendix's operand stack truncation. There has never been divergence between the two approaches until now, with the introduction of subtyping.

In the following example, the `br 0` makes the subsequent code unreachable. The appendix will additionally truncate the stack. Next, the `br_on_null` is typed `[anyref anyref] -> [anyref (ref any)]` due to the type of its label. With `wasm-smith`'s actual stack, this type checks fine because `structref <: anyref` combined with the rules for subtyping between instructions. For the appendix algorithm, we have an empty stack in unreachable code, so when the `br_on_null` wants to pop the `anyref` that the label demands, it does, and then when it pushes it back onto the stack, it pushes an `anyref`. Not a `structref`. This is the source of the divergence between the two approaches. Finally, the `br_on_cast_fail` observes the "actual" `structref`, triggering a `type mismatch: expected structref, found anyref` error in the appendix algorithm.

```wat
(module
  (func (param anyref) (result anyref)
    ;; 0x1b: appendix: []
    ;;         actual: []
    ref.null struct
    ;; 0x1d: appendix: [structref]
    ;;         actual: [structref]
    local.get 0
    ;; 0x1f: appendix: [structref anyref]
    ;;         actual: [structref anyref]
    br 0
    ;; 0x21: appendix: []
    ;;         actual: [structref]
    local.get 0
    ;; 0x23: appendix: [anyref]
    ;;         actual: [structref anyref]
    br_on_null 0
    ;; 0x25: appendix: [anyref (ref any)]
    ;;         actual: [structref (ref any)]
    drop
    ;; 0x26: appendix: [anyref]
    ;;         actual: [structref]
    br_on_cast_fail 0  structref structref
  )
)
```

The reference interpreter, and `wasmparser`, report the type mismatch error and consider the example invalid. But despite that, it isn't clear to me that it really is invalid.

Firefox and `wasm-smith` consider the example valid.

[...]

PR #518, opened 2024-02-07
Add a test for #516 to the test suite

fitzgen:
r? @rossberg

### DIFF
--- a/test/core/br_on_null.wast
+++ b/test/core/br_on_null.wast
@@ -64,3 +64,46 @@
 
 (assert_return (invoke "args-null" (i32.const 3)) (i32.const 3))
 (assert_return (invoke "args-f" (i32.const 3)) (i32.const 9))
+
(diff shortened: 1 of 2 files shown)


rossberg (commented):
Thanks! But since the issue it tests actually is the typing of br_on_null, I think it would be more appropriate to add  this test to br_on_null.wast.

fitzgen:
There isn't a `br_on_null.wast` in the `gc` directory. Should I create one?

rossberg:
It's in the main directory. But you'll need to modify the example to avoid the use of GC features (which should be inessential to the issue). E.g. with some use of subtyping that works with just the funcref proposal.

fitzgen:
@rossberg rewrote it to use `(ref null $ty)` and `funcref` instead of `structref` and `anyref` and moved the test into `br_on_null.wast`.

rossberg (approved):
Thanks!

Issue #501, opened 2023-12-20
[js-api] DefaultValue needs updating


rossberg:
The [DefaultValue algorithm](https://webassembly.github.io/gc/js-api/index.html#defaultvalue) needs updating to handle more general reference types. Moreover, it is partial now, so all use sites need to adapt to the possibility of it failing.

@takikawa

Also, this algorithm is unrelated to JS and in fact duplicates the respective [definition in the core spec](https://webassembly.github.io/gc/core/exec/runtime.html#default-val). It might be preferable to expose that in the core spec's [embedding interface](https://webassembly.github.io/gc/core/appendix/embedding.html) and call into that.

eqrion:
> Also, this algorithm is unrelated to JS and in fact duplicates the respective [definition in the core spec](https://webassembly.github.io/gc/core/exec/runtime.html#default-val). It might be preferable to expose that in the core spec's [embedding interface](https://webassembly.github.io/gc/core/appendix/embedding.html) and call into that.

The JS-API default value function is slightly different than wasm's default value function. `JSAPI-DefaultValue(externref) = undefined`, while `Wasm-DefaultValue(externref) = null`. This is thanks to JS filling in missing parameters with `undefined`.

rossberg:
Fair enough, but it would still be preferable to handle only this special case here and then defer to the core spec for all regular cases.

(Although personally I still think this special case is odd and instead it should only use undefined if explicitly given, not as default. There is precedence in JS for treating an absent parameter differently from explicit undefined, e.g., `[].reduce(() => 0)` throws TypeError, whereas `[].reduce(() => 0, undefined)` returns `undefined`.)

PR #522, opened 2024-02-15
[js-api][test] Refine DefaultValue for new reference types

takikawa:
This PR addresses issue #501 and adjust `DefaultValue` in the JS API so that it delegates to a embedder API function in the core spec, with a special case for `externref` to use `undefined`.

Note that for `anyref` and other non-externref types it will use `null` as the default value. For non-null types,  it produces `error` and that is thrown as a `TypeError` for `WebAssembly.Table.prototype.set` and `WebAssembly.Table.prototype.grow`.

Rendered spec change for the core spec:

image

Rendered change for `DefaultValue` (there are changes to use sites too, not shown):

image

The PR also adds a new WPT test for the error cases specifically. Both Firefox and Chrome passed the test when I tried it out.

### DIFF
--- a/document/core/appendix/embedding.rst
+++ b/document/core/appendix/embedding.rst
@@ -693,3 +693,23 @@ Matching
    \F{match\_externtype}(\X{et}_1, \X{et}_2) &=& \TRUE && (\iff \vdashexterntypematch \X{et}_1 \matchesexterntype \X{et}_2) \\
    \F{match\_externtype}(\X{et}_1, \X{et}_2) &=& \FALSE && (\otherwise) \\
    \end{array}
+
+
+.. index:: value type
+.. _embed-default-value:
+
+Value types
+~~~~~~~~~~~
+
+:math:`\F{default\_value}(\type) : \val`
+...............................................
+
+1. If :math:`\default_{type}` is not defined, then return :math:`\ERROR`.
+
+1. Else, return the :ref:`value <syntax-val>` :math:`\default_{type}`.
+
+.. math::
+   \begin{array}{lclll}
+   \F{default\_value}(t) &=& v && (\iff \default_t = v) \\
+   \F{default\_value}(t) &=& \ERROR && (\iff \default_t = \epsilon) \\
+   \end{array}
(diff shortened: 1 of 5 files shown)


document/core/appendix/embedding.rst
```diff
@@ -693,3 +693,23 @@ Matching
    \F{match\_externtype}(\X{et}_1, \X{et}_2) &=& \TRUE && (\iff \vdashexterntypematch \X{et}_1 \matchesexterntype \X{et}_2) \\
    \F{match\_externtype}(\X{et}_1, \X{et}_2) &=& \FALSE && (\otherwise) \\
    \end{array}
+
+
+.. index:: value type
```
rossberg, line 698:
```suggestion
.. index:: value type, value
```

document/core/appendix/embedding.rst
```diff
@@ -693,3 +693,23 @@ Matching
...
+.. _embed-default-value:
+
+Value types
+~~~~~~~~~~~
+
+:math:`\F{default\_value}(\type) : \val`
```
rossberg, line 704: For consistency with the naming conventions of the other functions, I would probably call this `val_default`, as a sort of value constructor. I'd also suggest to put it in the same section as `ref_type` (and rename the section to "Values").

rossberg (commented):
Excellent, thanks!

takikawa:
Thanks for the suggestions! Should be updated now:

image
image

document/core/appendix/embedding.rst
```diff
@@ -656,6 +657,20 @@ References
    In such cases, this function may return a less precise supertype.
 
 
+:math:`\F{val\_default}(\type) : \val`
+...............................................
+
+1. If :math:`\default_{type}` is not defined, then return :math:`\ERROR`.
+
+1. Else, return the :ref:`value <syntax-val>` :math:`\default_{type}`.
```
rossberg, line 665:
```suggestion
:math:`\F{val\_default}(\valtype) : \val`
...............................................

1. If :math:`\default_{valtype}` is not defined, then return :math:`\ERROR`.

1. Else, return the :ref:`value <syntax-val>` :math:`\default_{valtype}`.
```

rossberg (approved):
Thanks!

PR #523, opened 2024-02-15
[js-api] Throw consistent exns in "read the imports"

### LINKED ISSUE #521: [js-api] Throw LinkError consistently
The read-the-imports function delegates some cases to ToWebAssemblyValue, which can throw TypeError. But it is expected that all link-time errors produce LinkError. (cf. #498)

takikawa:
Fixes #521. Goes along with #498.

### DIFF
--- a/document/js-api/index.bs
+++ b/document/js-api/index.bs
@@ -370,7 +370,7 @@ A {{Module}} object represents a single WebAssembly module. Each {{Module}} obje
                     1. Throw a {{LinkError}} exception.
                 1. If |valtype| is [=v128=],
                     1. Throw a {{LinkError}} exception.
-                1. Let |value| be [=ToWebAssemblyValue=](|v|, |valtype|). If this operation throws an exception, catch it, and throw a {{LinkError}} exception.
+                1. Let |value| be [=ToWebAssemblyValue=](|v|, |valtype|). If this operation throws a {{TypeError}}, catch it, and throw a {{LinkError}} exception.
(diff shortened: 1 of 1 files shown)


document/js-api/index.bs
```diff
@@ -370,7 +370,7 @@ A {{Module}} object represents a single WebAssembly module. Each {{Module}} obje
                     1. Throw a {{LinkError}} exception.
                 1. If |valtype| is [=v128=],
                     1. Throw a {{LinkError}} exception.
-                1. Let |value| be [=ToWebAssemblyValue=](|v|, |valtype|).
+                1. Let |value| be [=ToWebAssemblyValue=](|v|, |valtype|). If this operation throws an exception, catch it, and throw a {{LinkError}} exception.
```
rossberg, line 373: I wonder if we should be specific to TypeError here. If there is any chance that TWAV could throw another exception (now or in the future), should that also become a LinkError? (Probably not.)

rossberg (approved):

document/js-api/index.bs
takikawa: I think it could make sense to be specific to TypeError, so I pushed that change. I think in practice it shouldn't be possible for TWAV to throw other exceptions at this particular call site currently.

For example, TWAV can throw a SyntaxError if BigInt conversion fails on a string but it should be guarded by the conditions here. But that property isn't immediately apparent either.

rossberg (approved):

PR #530, opened 2024-03-06
Fix typing of `ref.i31` in the appendix

fitzgen:
It produces a non-null `(ref i31)` not an `i31ref` aka `(ref null i31)`.


rossberg (approved):
Thanks!

Issue #201, opened 2021-03-03
Output type of local.tee?


askeksa-google:
With the introduction of static subtyping, the input to a `local.tee` can be a subtype of the type of the local. This gives two obvious, sound choices for the output type:

1. The type of the local. This is what follows from the typing rules for `local.tee` (same input, output and local type) combined with implicit upcasting of inputs.
2. The actual type of the input. This is more useful, since it is more precise. On the other hand, it means that `local.tee` is not equivalent to `local.set` followed by `local.get`, and doing the substitution can break validation.

Currently, as far as I can tell, Binaryen implements (1) and V8 implements (2).

What is the intended behavior?

kripken:
Previous discussion: https://github.com/WebAssembly/reference-types/issues/55

Looks like the status there is that we agreed to keep things as what you call option 1.

jakobkummerow:
OK, we'll fix that in V8 then.

(Though IMHO it's sad to discard perfectly valid type information, e.g. about non-null-ness in a sequence like `(struct.new) (local.tee) (struct.set)`...)

askeksa-google:
Thanks for the pointer.

This ties in with the discussion of non-nullable locals in https://github.com/WebAssembly/function-references/issues/44, in that when locals can't be non-nullable, any value stored through `local.tee` is forced to discard its nullability, leading to more inserted `ref.as_non_null` checks in practice.

kripken:
I'm not opposed to revisiting this topic, btw, I was just pointing out the previous info. Data on code size downsides could be important, I agree.

taralx:
At the time we were concerned about losing the equivalence. Do we have any data on that equivalence being used?

kripken:
I explored this in Binaryen, https://github.com/WebAssembly/binaryen/issues/3704#issuecomment-802291629 Overall there is some work to be done in the optimizer there to handle losing that equivalence, but it seems practical.

Now that we have an almost-fully working pipeline I think it's easier to make a decision on this. I'd support using the more specific type of the value, so my previous concerns are no longer valid. Perhaps we can make the change then, maybe in the next GC prototype version?

kripken:
Noticed in https://github.com/WebAssembly/binaryen/issues/3767 that `br_if` has a similar situation, where the spec says the type is that of the target block/loop/etc. and not the value flowing through it, which means that like with `local.tee` we are missing out on more specific type info.

[...]

PR #532, opened 2024-03-15
[Draft][spec] Improve typing of `br_if`

tlively:
`br_if` previously had type `[t* i32] -> [t*]` where `[t*]` was the type of its target label. This typing unnecessarily lost type information in cases where the actual input result type is a strict subtype of the label result type, requiring casts to recover the types that the validation algorithm already knew about before the `br_if`.

Update the type of `br_if` to be `[t1* i32] -> [t1*]` where `[t*]` is a subtype of the label type `[t*]`. This type preserves types that were present before the `br_if`, even when they are strict subsets of the label result types.

Update the description of principle types in the appendix to allow type variables to be constrained by upper bounds and update the validation algorithm to reflect the new typing.

This current PR illustrates the intended changes only for `br_if`, but the full intended change would similarly improve the types of `br_on_null`, `br_on_non_null`, `br_on_cast`, `br_on_cast_fail`, and `local.tee`.

Procedurally, we would want to run this by the full CG before making this change since the proposal is already phase 4. This is a non-breaking change, though, and until very recently was actually how most implementations behaved.

### DIFF
--- a/document/core/appendix/algorithm.rst
+++ b/document/core/appendix/algorithm.rst
@@ -380,8 +380,8 @@ Other instructions are checked in a similar manner.
        case (br_if n)
          error_if(ctrls.size() < n)
          pop_val(I32)
-         pop_vals(label_types(ctrls[n]))
-         push_vals(label_types(ctrls[n]))
+         let vts = pop_vals(label_types(ctrls[n]))
+         push_vals(vts)
 
        case (br_table n* m)
          pop_val(I32)
--- a/document/core/appendix/properties.rst
+++ b/document/core/appendix/properties.rst
@@ -1344,7 +1344,7 @@ The :ref:`type system <type-system>` of WebAssembly features both :ref:`subtypin
 That has the effect that every instruction or instruction sequence can be classified with multiple different instruction types.
 
 However, the typing rules still allow deriving *principal types* for instruction sequences.
-That is, every valid instruction sequence has one particular type scheme, possibly containing some unconstrained place holder *type variables*, that is a subtype of all its valid instruction types, after substituting its type variables with suitable specific types.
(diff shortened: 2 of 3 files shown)


conrad-watt:
> Update the description of principle types in the appendix to allow type variables to be constrained by upper bounds and update the validation algorithm to reflect the new typing.

Without more justification, I'd be against this. It seems like an unfortunate last-minute complication to our conceptual model of how typing works to allow these kinds of bounds. Is the aim here to bless what some implementations are doing, due to the divergences in https://github.com/WebAssembly/gc/issues/516?

tlively:
> Is the aim here to bless what some implementations are doing, due to the divergences in https://github.com/WebAssembly/gc/issues/516?

No, not directly, but it is related. The aim is to improve the precision of the types for these instructions so we admit programs with fewer casts than before. The reason implementations were wrong in the first place is that the implementers incorrectly assumed the spec would already be giving these instructions the most precise possible types for the same reasons. From the implementer point of view, this is fixing a bug in the spec, not just changing the spec to match implementations.

conrad-watt:
> The aim is to improve the precision of the types for these instructions so we admit programs with fewer casts than before.

> From the implementer point of view, this is fixing a bug in the spec, not just changing the spec to match implementations.

I don't really agree with these points. We spent a lot of time debating the exact principal types property we wanted, and how we were going to type `local.tee` [specifically](https://github.com/WebAssembly/gc/issues/201). I think the presumption should be that we stick with these decisions rather than reopen them due to an implementation discrepancy. I appreciate that the issue with `br_*` seems to specifically create more work for Binaryen because of the inference vs checking distinction, but it seems like there's a feasible patch? (https://github.com/WebAssembly/binaryen/pull/6390).

We did note in https://github.com/WebAssembly/gc/issues/201 that we could potentially revisit the discussion later, but we should acknowledge that this patch is effectively reopening the https://github.com/WebAssembly/gc/issues/201 discussion, rather than making an uncontroversial improvement.

tlively:
Yes, my intent is to reopen the discussion from #201 (and https://github.com/WebAssembly/reference-types/issues/55). Sorry if that wasn't clear.

PR #534, opened 2024-04-04
Add additional `i31ref` spec tests

fitzgen:
I noticed that Wasmtime was passing the spec tests despite having bits that were known to be unimplemented. This should help the tests exercise those corners of the spec.

Notably:

* Setting `i31ref` globals.

* Initializing tables and globals with `(ref.i31 (global.get $g))`.

* Table operations on `i31ref` tables.

* Accessing `anyref` globals and tables that are actually `i31ref`s. This is interesting to exercise outside of general subtyping because we have different paths in our inline GC barriers for `anyref`s that are actually `i31ref`s.

@rossberg PTAL

### DIFF
--- a/test/core/gc/i31.wast
+++ b/test/core/gc/i31.wast
@@ -26,9 +26,7 @@
   )
 
   (func (export "set_global") (param i32)
-    local.get 0
-    ref.i31
-    global.set $m
+    (global.set $m (ref.i31 (local.get 0)))
   )
 )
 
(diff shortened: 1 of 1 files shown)


rossberg (commented):
Thanks, this looks good. Should we also have non-nullable variants of these tests?

One nit: Can you match the folded style of the rest of this test?

fitzgen:
> Should we also have non-nullable variants of these tests?

There are a couple non-null types in there. Indeed, we avoid generating null-checks for GC barriers when we have non-null types. I think in an ideal world, yes, we would duplicate ~everything once for nullable and once more for non-nullable types. Don't really want to do that by hand though. Maybe if we were generating tests programmatically.



> One nit: Can you match the folded style of the rest of this test?

Done!

rossberg (approved):

fitzgen:
Thanks!

PR #537, opened 2024-04-15
[Spec] Fix bounds check prose for array.fill and array.copy

### LINKED ISSUE #536: Bounds check is possible wrong on array.copy/array.fill
Apologies I'm currently implementing this spec so I you might see few issues/questions raised from me

Both array.copy and array fill have a bounds checks  that effectively equate to:

if offset + elements_to_copy >= array.len => trap

I believe these should all be > rather than >=, I realised this when testing (because I encountered traps where I didn't expect to) and to double check I compared against table.copy/table.fill and those specifications use > rather than >=

tlively:
The prose incorrectly used "larger than or equal to" when it should have used just "larger than." The formalism was already correct.

Fixes #536.


rossberg (approved):
Thanks!

PR #542, opened 2024-05-01
Change monomorphisation to monomorphization

OsamaAhmad00:
Though monomorphisation is correct, it's referenced as monomorphization in most of the references.


rossberg:
Well, the whole doc is currently spelled in British English, with various other occurrences of "s" vs "z" and the likes (sometimes in the same sentence). Only converting some of them into American English would be a bit inconsistent. ;)

PR #548, opened 2024-07-05
Fix `RETURNCALLREF` argument type

eduVcJy0UE0cy0zo3BS7XYO3b:
According to other places in docs and my research on dumped wat files, I consider that funcidx for this call places on stack, not as parameter.


rossberg (approved):
Indeed, thanks!

PR #550, opened 2024-07-10
Add tests that call_ref does not accept funcref

tlively:


rossberg (approved):
If we wanted to be pedantic, this PR should be on the function-refs repo. But at this point it doesn't really matter anymore. :)

PR #551, opened 2024-07-24
[spec] Fix execution semantics prose of `ARRAY.NEW_DATA`

f52985:
This PR fixes the execution semantics prose of `ARRAY.NEW_DATA`,
which was using a wrong iteration variable for the byte-subsequence `b'`.


rossberg (approved):
Thanks!

PR #562, opened 2024-09-27
Expand tests for `array.new_data`

fitzgen:
I noticed that the existing tests didn't exercise out-of-bounds indexing of the data segment, so I added a test for that.

At the same time, since I was already creating a dedicated `.wast` file for testing `array.new_data`, I also added a few general tests to round the file out. Some of these might technically duplicate things that already happen to be tested in `array.wast`, but I think that's probably fine, and that it is good to have dedicated tests in this new file. I can remove them if we'd rather not have these potential-duplicate tests.

### DIFF
--- a/test/core/gc/array_new_data.wast
+++ b/test/core/gc/array_new_data.wast
@@ -1,0 +1,50 @@
+(module
+  (type $arr (array (mut i8)))
+
+  (data $d "abcd")
+
+  (func (export "array-new-data") (param i32 i32) (result (ref $arr))
+    (array.new_data $arr $d (local.get 0) (local.get 1))
+  )
+)
+
+;; In-bounds data segment accesses.
+(assert_return (invoke "array-new-data" (i32.const 0) (i32.const 0)) (ref.array))
+(assert_return (invoke "array-new-data" (i32.const 0) (i32.const 4)) (ref.array))
+(assert_return (invoke "array-new-data" (i32.const 1) (i32.const 2)) (ref.array))
+(assert_return (invoke "array-new-data" (i32.const 4) (i32.const 0)) (ref.array))
+
+;; Out-of-bounds data segment accesses.
+(assert_trap (invoke "array-new-data" (i32.const 0) (i32.const 5)) "out of bounds memory access")
+(assert_trap (invoke "array-new-data" (i32.const 5) (i32.const 0)) "out of bounds memory access")
+
+(module
+  (type $arr (array (mut i8)))
+
+  (data $d "\aa\bb\cc\dd")
+
+  (func (export "array-new-data-contents") (result i32 i32)
+    (local (ref null $arr))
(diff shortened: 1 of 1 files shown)


test/core/gc/array_new_data.wast
```diff
@@ -0,0 +1,50 @@
...
+
+;; In-bounds data segment accesses.
+(assert_return (invoke "array-new-data" (i32.const 0) (i32.const 0)) (ref.array))
+(assert_return (invoke "array-new-data" (i32.const 0) (i32.const 4)) (ref.array))
+(assert_return (invoke "array-new-data" (i32.const 1) (i32.const 2)) (ref.array))
+(assert_return (invoke "array-new-data" (i32.const 4) (i32.const 0)) (ref.array))
+
+;; Out-of-bounds data segment accesses.
+(assert_trap (invoke "array-new-data" (i32.const 0) (i32.const 5)) "out of bounds memory access")
+(assert_trap (invoke "array-new-data" (i32.const 5) (i32.const 0)) "out of bounds memory access")
```
rossberg, line 19: Perhaps also add
```
(assert_trap (invoke "array-new-data" (i32.const 4) (i32.const 1)) "out of bounds memory access")
```

test/core/gc/array_new_data.wast
```diff
@@ -0,0 +1,50 @@
...
+
+(module
+  (type $arr (array (mut i8)))
+
+  (data $d "\aa\bb\cc\dd")
+
+  (func (export "array-new-data-contents") (result i32 i32)
+    (local (ref null $arr))
```
rossberg, line 27:
```suggestion
    (local (ref $arr))
```

test/core/gc/array_new_data.wast
```diff
@@ -0,0 +1,50 @@
...
+
+  (func (export "array-new-data-little-endian") (result i32)
+    (array.get $arr
+               (array.new_data $arr $d (i32.const 0) (i32.const 1))
+               (i32.const 0))
+  )
+)
+
+;; Data segments are interpreted as little-endian.
+(assert_return (invoke "array-new-data-little-endian") (i32.const 0xddccbbaa))
```
rossberg, line 50: Perhaps extend this test with an example where the originating i32 value is not word-aligned in the data segment.

rossberg (approved):
LGTM

test/core/gc/array_new_data.wast
fitzgen: Added this and `(invoke "array-new-data" (i32.const 1) (i32.const 4))`.

test/core/gc/array_new_data.wast
fitzgen: Done

test/core/gc/array_new_data.wast
fitzgen: Added a test for unaligned data.

fitzgen:
Thanks for review, I've pushed updates.

PR #566, opened 2024-10-25
Fix test copy/paste typo

ejrgilbert:


tlively:
Looks right, but curious that the tests weren't previously failing.

jakobkummerow:
> curious that the tests weren't previously failing.

`i31.get_s` and `i31.get_u` both trap on `(ref.null i31)`, so this doesn't affect test results.

jakobkummerow (approved):
LGTM

tlively:
Ah, I misread and though that this function was used to test non-null cases.

PR #573, opened 2024-12-03
Add some tests for invalid subtyping due to non-matching field types

fitzgen:
Specifically around the requirements that both field types have the same mutability and that the storage type must be exactly the same when the field type is mutable.

@tlively PTAL, thanks!

### DIFF
--- a/test/core/gc/type-subtyping-invalid.wast
+++ b/test/core/gc/type-subtyping-invalid.wast
@@ -19,19 +19,19 @@
  )
 
 (assert_invalid
- (module
-   ;; The mutability of fields must be the same.
-   (type $c (sub    (struct (field (mut (ref null any))))))
-   (type $d (sub $c (struct (field      (ref null none)))))
+  (module
+    ;; The mutability of fields must be the same.
+    (type $c (sub    (struct (field (mut (ref null any))))))
+    (type $d (sub $c (struct (field      (ref null any)))))
   )
   "sub type 1 does not match super type"
 )
 
 (assert_invalid
- (module
-   ;; The mutability of fields must be the same.
-   (type $c (sub    (struct (field      (ref null any)))))
-   (type $d (sub $c (struct (field (mut (ref null none))))))
+  (module
+    ;; The mutability of fields must be the same.
+    (type $c (sub    (struct (field      (ref null any)))))
+    (type $d (sub $c (struct (field (mut (ref null any))))))
   )
   "sub type 1 does not match super type"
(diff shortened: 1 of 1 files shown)


test/core/gc/type-subtyping-invalid.wast
```diff
@@ -0,0 +1,37 @@
...
+    (type $b (sub $a (struct (field (mut (ref null any))))))
+  )
+  "sub type 1 does not match super type"
+ )
+
+(assert_invalid
+ (module
+   ;; The mutability of fields must be the same.
+   (type $c (sub    (struct (field (mut (ref null any))))))
+   (type $d (sub $c (struct (field      (ref null none)))))
```
rossberg, line 25: Perhaps keep the actual storage types the same in this test (and the next one)?

rossberg (commented):
Thanks! Why not include these in the existing type-subtyping?

test/core/gc/type-subtyping-invalid.wast
fitzgen: > Perhaps keep the actual storage types the same in this test (and the next one)?

Whoops, yeah that was the intention, just a copy-paste issue. Will push a fix in a moment.

test/core/gc/type-subtyping-invalid.wast
```diff
@@ -0,0 +1,37 @@
...
+  )
+  "sub type 1 does not match super type"
+)
+
+(assert_invalid
+  (module
+    ;; When fields are const, a subtype's reference fields cannot be supertypes of
+    ;; the supertype's fields, they must be subtypes.
+    (type $a (sub    (struct (field (mut (ref null none))))))
+    (type $b (sub $a (struct (field (mut (ref null any))))))
```
tlively, line 16: It looks like this test intends to have constant fields, but in fact it has mutable fields. Do we already test the case where the subtyping is correctly covariant, but the fields are mutable so it is still invalid?

tlively (approved):

fitzgen:
Thanks for the approve @tlively! FWIW, I do not have merge permissions on this repo, so I can't merge this PR myself.

rossberg:
I regularly go through the proposal repos and merge them into wasm-3.0, so new commits should eventually make it there. I haven't done any back-merging, because that is way to tedious, and at this point I'm not sure how much it still matters.

tlively:
Sounds good, I'll go ahead and merge this, then.

PR #581, opened 2025-04-22
small typo in Makefile

evancz:
Change directory path `test/corel` to `test/core`


rossberg (approved):
Huh, that's odd... thanks!

